PR #28, opened 2021-11-05
Set SignalInstances directly as attributes on objects (fix bug with hashable signal holders)

### LINKED ISSUE #27: bug: Unhashable types cannot have signals
```py
In [14]: from psygnal import Signal

In [15]: class T(dict):
    ...:     s = Signal()
    ...:

In [17]: t= T()

In [18]: t.s
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-18-10c22990d292> in <module>
----> 1 t.s

~/miniconda3/envs/qt/lib/python3.9/site-packages/psygnal/_signal.cpython-39-darwin.so in psygnal._signal.Signal.__get__()

~/miniconda3/envs/qt/lib/python3.9/weakref.py in setdefault(self, key, default)
    507
    508     def setdefault(self, key, default=None):
--> 509         return self.data.setdefault(ref(key, self._remove),default)
    510
    511     def update(self, dict=None, /, **kwargs):

TypeError: unhashable type: 'T'
```

cc @Czaki

possible solutions:

1. don't use weakkeydict to store signal instance (use `id(obj)` and `finalize(pop...)` or something)
2. possibly just kill the descriptor on first attribute access? and `setattr(obj, name, SignalInstance(...))`?

tlambert03:
fixes #27 

not sure if this will break other (unanticipated) things, so going to let this sit for a bit

h/t @Czaki

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,4 +35,4 @@ benchmark-all:
 
 # compare HEAD against main
 benchmark-compare:
-	asv compare --split --factor 1.1 --only-changed main HEAD
+	asv compare --split --factor 1.1 main HEAD
--- a/psygnal/_signal.py
+++ b/psygnal/_signal.py
@@ -206,8 +206,9 @@ class Signal:
         # to instance.name ... this essentially breaks the descriptor,
         # (i.e. __get__ will never again be called for this instance, and we have no
         # idea how many instances are out there),
-        # but it allows us to prevent creating a key for this instance,
-        # which may not be hashable or weak-referenceable.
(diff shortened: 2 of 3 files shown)


tlambert03:
wow, @czaki, something I didn't expect here (but makes sense when you think about it) is the big performance gain!

https://github.com/tlambert03/psygnal/runs/4126665312?check_suite_focus=true#step:5:117

presumably faster attribute access leading to general speedup

PR #29, opened 2021-11-05
Avoid holding strong reference to decorated and partial methods

Czaki:
In this PR there is fix for selection of wrong select (most common case is decorator which not mock `__name__`)

Also partial mathods are unwrapped and weakref to base object is keep.

_edit_: ... an example of things being fixed here:

```py
def deco(fun):
    def _fun(*args):
        fun(*args)
    return _fun

class MyObj:
    @deco
    def method(self, a: int): ...
    assigned_func = lambda self, a: None

class Emitter:
    changed = Signal(int)

e = Emitter()
obj = MyObj()

# these would have failed
e.changed.connect(obj.method)
e.changed.connect(obj.assigned_func)

# this would have retained a strong reference
e.changed.connect(partial(obj.method))
```

### DIFF
--- a/psygnal/_signal.py
+++ b/psygnal/_signal.py
@@ -6,7 +6,8 @@ import warnings
 import weakref
 from contextlib import contextmanager
 from functools import lru_cache, partial, reduce
-from inspect import Parameter, Signature, ismethod
+from inspect import Parameter, Signature
+from types import MethodType
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -21,9 +22,7 @@ from typing import (
     Union,
     cast,
     overload,
-    Protocol,
 )
-from types import MethodType
 
 from typing_extensions import Literal
 
@@ -40,15 +39,14 @@ class PartialMeta(type):
     def __instancecheck__(cls, inst: object) -> bool:
         return isinstance(inst, partial) and isinstance(inst.func, MethodType)
 
+
 class Partial(metaclass=PartialMeta):
     func: MethodType
     args: List[Any]
     keywords: Dict
 
-    def __call__(self, *args: List[Any], **kwargs:Dict[str, Any]) -> Any:
+    def __call__(self, *args: List[Any], **kwargs: Dict[str, Any]) -> Any:
         raise NotImplementedError
-
-
 
 
 def signature(obj: Any) -> inspect.Signature:
--- a/psygnal/_signal.py
+++ b/psygnal/_signal.py
@@ -5,7 +5,7 @@ import threading
 import warnings
 import weakref
 from contextlib import contextmanager
-from functools import lru_cache, reduce, wraps, partial
+from functools import lru_cache, partial, reduce
 from inspect import Parameter, Signature, ismethod
 from typing import (
     TYPE_CHECKING,
@@ -667,7 +667,9 @@ class SignalInstance:
                         if cb is None:  # pragma: no cover
                             rem.append(slot)  # object has changed?
                             continue
-                    elif hasattr(slot, "__ref__") and slot.__ref__() is None:
+                    elif (
+                        hasattr(slot, "__ref__") and slot.__ref__() is None  # type: ignore  # noqa: E501
+                    ):
                         rem.append(slot)  # add dead weakref
                         continue
                     else:
@@ -919,31 +921,29 @@ def _is_subclass(left: AnyType, right: type) -> bool:
     return issubclass(left, right)
 
 
-def partial_weakref(partial_fun):
+def partial_weakref(partial_fun: partial) -> Callable:
     obj, name = _get_proper_name(partial_fun.func)
     args_ = partial_fun.args
     kwargs_ = partial_fun.keywords
-    def wrap(*args, **kwargs):
+
+    def wrap(*args, **kwargs):  # type: ignore
         ob = obj()
         if ob is None:
             return
         getattr(ob, name)(*args_, *args, **kwargs_, **kwargs)
 
-    wrap.__ref__ = obj
-    # wrap.__name__ = partial_fun.__name__
-    # wrap.__annotations__ = partial_fun.__annotations__
-    # wrap.__doc__ = partial_fun.__doc__
+    wrap.__ref__ = obj  # type: ignore
 
     del partial_fun
     return wrap
 
 
-def _get_proper_name(callback):
+def _get_proper_name(callback: Callable) -> Tuple[weakref.ref, str]:
     assert inspect.ismethod(callback)
-    obj = callback.__self__
+    obj = callback.__self__  # type: ignore
     if (
-            not hasattr(obj, callback.__name__)
-            or getattr(obj, callback.__name__) != callback
+        not hasattr(obj, callback.__name__)
+        or getattr(obj, callback.__name__) != callback
     ):
         # some decorators will alter method.__name__, so that obj.method
         # will not be equal to getattr(obj, obj.method.__name__). We check
--- a/psygnal/_signal.py
+++ b/psygnal/_signal.py
@@ -5,7 +5,7 @@ import threading
 import warnings
 import weakref
 from contextlib import contextmanager
-from functools import lru_cache, reduce
+from functools import lru_cache, reduce, wraps, partial
 from inspect import Parameter, Signature, ismethod
 from typing import (
     TYPE_CHECKING,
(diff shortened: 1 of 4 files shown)


Czaki:
Maybe type annotation could be improved.

tlambert03:
> Maybe type annotation could be improved.

elaborate?

psygnal/_signal.py
```diff
@@ -471,7 +471,9 @@ def _raise_connection_error(self, slot: Callable, extra: str = "") -> NoReturn:
 
     def _normalize_slot(self, slot: NormedCallback) -> NormedCallback:
         if ismethod(slot):
-            return (weakref.ref(slot.__self__), slot.__name__)  # type: ignore
+            return _get_proper_name(slot)  # type: ignore
```
Czaki, line 474: At this place, I do not know if I could do something better than ignore warning.

psygnal/_signal.py
```diff
@@ -912,3 +919,40 @@ def _is_subclass(left: AnyType, right: type) -> bool:
...
             return any(issubclass(i, right) for i in get_args(left))
     return issubclass(left, right)
+
+
+def partial_weakref(partial_fun: partial) -> Callable:
+    obj, name = _get_proper_name(partial_fun.func)
+    args_ = partial_fun.args
+    kwargs_ = partial_fun.keywords
+
+    def wrap(*args, **kwargs):  # type: ignore
```
Czaki, line 929: I have no idea if it could be annotated in any way.

psygnal/_signal.py
```diff
@@ -912,3 +919,40 @@ def _is_subclass(left: AnyType, right: type) -> bool:
...
+
+    wrap.__ref__ = obj  # type: ignore
+
+    del partial_fun
+    return wrap
+
+
+def _get_proper_name(callback: Callable) -> Tuple[weakref.ref, str]:
+    assert inspect.ismethod(callback)
+    obj = callback.__self__  # type: ignore
```
Czaki, line 943: There is any option to annotate the function as a method (to remove this ignore)?

Czaki (commented):
I have pointed places which may be maybe improved?

psygnal/_signal.py
tlambert03: we could maybe use `types.MethodType` ... and instead of `inspect.ismethod` (Which apparently doesn't do the required type narrowing) we can use `isinstance(obj, MethodType)` ... then make the argument `callback: MethodType` instead of `Callable`?

psygnal/_signal.py
```diff
@@ -912,3 +919,40 @@ def _is_subclass(left: AnyType, right: type) -> bool:
...
+    args_ = partial_fun.args
+    kwargs_ = partial_fun.keywords
+
+    def wrap(*args, **kwargs):  # type: ignore
+        ob = obj()
+        if ob is None:
+            return
+        getattr(ob, name)(*args_, *args, **kwargs_, **kwargs)
+
+    wrap.__ref__ = obj  # type: ignore
```
tlambert03, line 935:
```suggestion
    setattr(wrap, '__ref__', obj)
```

avoids the ignore

psygnal/_signal.py
tlambert03:
```suggestion
    def wrap(*args: Any, **kwargs: Any) -> Any:
```

works for me

psygnal/_signal.py
```diff
@@ -471,7 +471,9 @@ def _raise_connection_error(self, slot: Callable, extra: str = "") -> NoReturn:
 
     def _normalize_slot(self, slot: NormedCallback) -> NormedCallback:
         if ismethod(slot):
```
tlambert03, line 473:
```suggestion
        if isinstance(slot, MethodType):
```

psygnal/_signal.py
```diff
@@ -912,3 +919,40 @@ def _is_subclass(left: AnyType, right: type) -> bool:
...
+        getattr(ob, name)(*args_, *args, **kwargs_, **kwargs)
+
+    wrap.__ref__ = obj  # type: ignore
+
+    del partial_fun
+    return wrap
+
+
+def _get_proper_name(callback: Callable) -> Tuple[weakref.ref, str]:
+    assert inspect.ismethod(callback)
```
tlambert03, line 942:
```suggestion
def _get_proper_name(callback: MethodType) -> Tuple[weakref.ref, str]:
```

psygnal/_signal.py
```diff
@@ -912,3 +919,40 @@ def _is_subclass(left: AnyType, right: type) -> bool:
         if origin is Union:
             return any(issubclass(i, right) for i in get_args(left))
     return issubclass(left, right)
+
+
+def partial_weakref(partial_fun: partial) -> Callable:
+    obj, name = _get_proper_name(partial_fun.func)
```
tlambert03, line 925:
```suggestion
    if not isinstance(partial_fun.func, MethodType):
        return partial_fun
    obj, name = _get_proper_name(partial_fun.func)
```

psygnal/_signal.py
```diff
@@ -471,7 +471,9 @@ def _raise_connection_error(self, slot: Callable, extra: str = "") -> NoReturn:
 
     def _normalize_slot(self, slot: NormedCallback) -> NormedCallback:
         if ismethod(slot):
-            return (weakref.ref(slot.__self__), slot.__name__)  # type: ignore
+            return _get_proper_name(slot)  # type: ignore
+        if isinstance(slot, partial) and ismethod(slot.func):
+            return partial_weakref(slot)
```
tlambert03, line 476:
```suggestion
        if isinstance(slot, partial):
            return partial_weakref(slot)
```

tlambert03 (commented):
some typing suggestions
(all need `from types import MethodType` at the top)

Czaki:
some things I solved in a different way.

tlambert03:
unfortunately, this one isn't as lucky on the performance side: https://github.com/tlambert03/psygnal/runs/4126896036?check_suite_focus=true#step:5:117

of all the methods here, I think `emit` is probably the one that we should try to keep as fast as possible, and this PR is currently hurting emit (particularly with a lot of connections).  Can you think of any tricks we might use to avoid that penalty?

Czaki:
it looks like the checking attribute is slow. 
Maybe a better option will be catching attribute errors instead of checking if the object is alive? Then it could be faster than the current implementation but with a bigger penalty for the dead object. What did you think?

tlambert03:
yeah, that's worth a try.  I'm playing with it a bit now.  The other thing we could try is just store it upfront in the StoredSlot tuple? that seems to be fast for the other stuff.   I can also tinker with it a bit since i've got the asv stuff setup to run locally at the moment.

Czaki:
>  I can also tinker with it a bit since i've got the asv stuff setup to run locally at the moment.

I do not have conda so I cannot run `asv` locally. 

The other option is to use wrapping class, not tuples. 

```
class Wrapper:
	def __init__(self, slot):
		self.obj, self.name = _get_proper_name(slot)
		
	def __cal__(self, *args, **kwargs):
		return getattr(self.obj(), self.name)(*args, **kwargs)
		
	def is_alive(self)
		return self.obj() is not None

class PartialWrapper(Wrapper):
	def __init__(self, slot):
		super().__init__(slot.func)
	    self.args = slot.args
    	self.kwargs = slot.keywords
    	
    def __call__(self, *args, **kwargs):
            getattr(self.obj, self.name)(*self.args, *args, **self.kwargs, **kwargs)
```

Then emit code could be simpler. 
What did you think?

tlambert03:
seems worth a try!  will report a bit more tomorrow

psygnal/_signal.py
```diff
@@ -913,3 +935,40 @@ def _is_subclass(left: AnyType, right: type) -> bool:
...
+        # will not be equal to getattr(obj, obj.method.__name__). We check
+        # for that case here and traverse to find the right method here.
+        for name in dir(obj):
+            meth = getattr(obj, name)
+            if inspect.ismethod(meth) and meth == callback:
+                return weakref.ref(obj), name
+        raise RuntimeError(
+            f"During bind method {callback} of object {obj} an error happen"
+        )
+    return weakref.ref(obj), callback.__name__
```
tlambert03, line 974:
```suggestion
def _get_proper_name(slot: MethodType) -> tuple[weakref.ref, str]:
    obj = slot.__self__
    # some decorators will alter method.__name__, so that obj.method
    # will not be equal to getattr(obj, obj.method.__name__). 
    # We check for that case here and find the proper name in the function's closures
    if getattr(obj, slot.__name__, None) != slot:
        for c in slot.__closure__ or ():
            cname = getattr(c.cell_contents, "__name__", None)
            if cname and getattr(obj, cname, None) == slot:
                return weakref.ref(obj), cname
        raise RuntimeError(
            f"Could not find method on {obj} corresponding to decorated function {slot}"
        )
    return weakref.ref(obj), slot.__name__
```

not critical here, since this is only during connection, but this is about 20x faster. (searching through the decorator's closures instead of `dir(class)`

Czaki: next nice thing which I do not know earlier.

Czaki:
using __closure__ does not help when using manual assignment to function. Maybe __closure__s should be used first but theh back to callback? 

```ipython
class A:
    def a(self):
        pass
    
A.b = lambda x: True 
a = A()
a.b.__self__
Out[20]: <__main__.A at 0x7f93282a67f0>
_get_proper_name(a.b)
Traceback (most recent call last):
  File "/home/czaki/.pyenv/versions/psygnal/lib/python3.9/site-packages/IPython/core/interactiveshell.py", line 3444, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-21-1e7d1b14854f>", line 1, in <module>
    _get_proper_name(a.b)
  File "/home/czaki/Projekty/psygnal/psygnal/_signal.py", line 966, in _get_proper_name
    raise RuntimeError(
RuntimeError: Could not find method on <__main__.A object at 0x7f93282a67f0> corresponding to decorated function <bound method <lambda> of <__main__.A object at 0x7f93282a67f0>>
```

tlambert03:
> using __closure__ does not help when using manual assignment to function.

great catch, thanks.

good news, we're back in the green on the benchmarks!

tlambert03:
> Maybe __closure__ should be used first but then back to callback?

took your advice and added a test.  This looks great to me now (two very tricky cases here 👍 thanks for adding these!).  I'll stop touching it in case you want to add or change anything else.  otherwise will merge.

tlambert03 (approved):

Czaki:
Looks ok for me

Issue #36, opened 2021-11-07
Idea: Allow conect setting attribute to Signal


Czaki:
As psygnal comes from Qt which is a C++ world that does not contain the equivalent of `@property` or descriptors. Using property is quite popular in python projects, but connected with signals could produce a big amount of redundant code. 

So maybe it could be a nice idea to allow connect a property to signal. 

```python
def property_weakref(obj: Union[weakref.ref, object], name:str) -> Tuple[weakref.ref, Callable]:
    if not isinstance(obj, weakref.ref):
        obj = weakref.ref(obj)
    def wrap(*args: Any) -> None:
        if len(args) == 1:
            setattr(obj(), name, args[0])
        else:
            setattr(obj(), name, args)

    return (obj, wrap)
```

and a proper modification for the `connect` function which allows providing tuple `(obj, str)` as input.

tlambert03:
love it.  dealing with setter callbacks is one of the most annoying things about using `property.setter`... would be great to make that case seamless

tlambert03:
(can't decide if i think it should be a new `connect` method though:

```py
emitter.connect_attr(obj, 'attr_name')
```

do you want to implement this?  Or just happy to see it get added?

PR #38, opened 2021-11-07
Connet set attribute to signal (alternate)

Czaki:
This PR implements #36

### DIFF
--- a/psygnal/_signal.py
+++ b/psygnal/_signal.py
@@ -17,18 +17,33 @@ from typing import (
     List,
     NoReturn,
     Optional,
+    Protocol,
     Tuple,
     Type,
     Union,
     cast,
     overload,
+    runtime_checkable,
 )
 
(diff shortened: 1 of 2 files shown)


psygnal/_signal.py
```diff
@@ -940,6 +1001,89 @@ def wrap(*args: Any, **kwargs: Any) -> Any:
...
+            setattr(self.obj(), self.name, args[0])
+        else:
+            setattr(self.obj(), self.name, args)
+
+    def __eq__(self, other: Any) -> bool:
+        return (
+            isinstance(other, PropertyWeakrefCallback)
+            and self.name == other.name
+            and self.obj() == other.obj()
+        )
```
Czaki: I decide to switch to class to allow compare two objects

tlambert03:
merged #39, so converted this to draft.  Some nice class patterns in here that we should remember.  perhaps we can figure out the performance stuff later and use it then

PR #42, opened 2021-11-07
Add SignalGroup

tlambert03:
SignalGroup that enables connecting to all SignalInstances.

```py
class MyGroup(SignalGroup):
    sig1 = Signal(int)
    sig2 = Signal(str)

events = MyGroup()

def my_callback(record):
    signal_name = record.signal.name
    args = record.args

events.connect(my_callback)
```

unlike `SignalInstance.connect`, callbacks connected to `SignalGroup.connect` will receive a single  `EmissionInfo` tuple argument which contains info about the signal instance the event came from:

 ```py
class EmissionInfo(NamedTuple):
    """Tuple containing information about an emission event."""

    signal: SignalInstance
    args: Tuple[Any, ...]
    extra_info: Dict[str, Any] = {}
```

if you'd rather get the args directly passed to the callback, you can use `SignalGroup.connect_direct`


tlambert03:
@alisterburt, resuscitating this.  If you have a quick moment to peek at the API, that'd be great

alisterburt:
API looks lovely, super clean as always! Did you have any concerns?

I don't know when I would be parsing an 'EmissionInfo' object for specific data, the use case I imagine is 'something changed, I want to know!' but having the data available is definitely nice!

tlambert03:
> Did you have any concerns?

not really, just makes me feel like someone's out there :)

> I don't know when I would be parsing an 'EmissionInfo' object for specific data, 

yeah, this was my way to deal with the difference between the vispy pattern (where callbacks always receive an event object) and here, where callbacks receive the args directly as in Qt.  When we have a ton of different signals, we can't necessarily connect a single callback to all of them, since the signature will vary.  So this normalizes it all.  it's fine to connect a callback that takes _no_ arguments (if all you want to know is that any of them have changed)... but if you did want to know which one changed, this lets you do it.

thanks!

alisterburt:
> not really, just makes me feel like someone's out there :)

Out here and loving what you're doing! 🙂

PR #49, opened 2021-11-17
support Qt SignalInstance Emit

### LINKED ISSUE #47: Chaining signal from psygnal to Qt Signals
When trying to propagate event from psygnal to Qt Signal (to use Qt benefit of invoking in proper thread for example) now you need to declare additional functions to do this. It will be nice to have code such this working:

```python
import sys

from qtpy.QtCore import Signal, QObject, QCoreApplication
from psygnal import Signal as PSignal

class SomePythonClass:
    some_inner_signal = PSignal(str)

    def call_signal(self):
        self.some_inner_signal.emit("test text")

class SampleQtClass(QObject):
    sample_signal = Signal()

    def __init__(self):
        super().__init__()
        self._inner_object = SomePythonClass()
        self._inner_object.some_inner_signal.connect(self.sample_signal.emit)

    def call_func(self):
        self._inner_object.call_signal()


def called():
    print("text")

if __name__ == '__main__':
    app = QCoreApplication(sys.argv)

    ob = SampleQtClass()
    ob.sample_signal.connect(called)

    ob.call_func()
```

At this moment I do not know option to inspect Qt Signal signature, but maybe good workaround will be add `maxargs` argument to `connect` function to allow user provide this information in code?

tlambert03:
closes #47 

@Czaki ... would be great if you could have a look, particularly at the tests


Czaki:
For me looks ok. but it looks like it is time to switch to `tox` or `nox` for tests.

tlambert03:
> For me looks ok. but it looks like it is time to switch to tox or nox for tests.

because of the backend switch?  yeah... that's not awesome... but I don't expect we'll be doing much of that (I really didn't want to have _any_ qt in here, but i get that it's important).  probably gonna leave this consideration for later though. 

thanks!

tlambert03:
we're taking a slight hit on connection time performance here,  not surprisingly.  It's just a few microseconds in most cases, so i think it's worth the gain.  going to merge even though benchmarks are failing

PR #56, opened 2022-02-06
Evented containers

tlambert03:
adding "evented" containers (like the ones in napari) backed by signal instead of vispy-like events


alisterburt (commented):
looooooove this - not having finished napari/superqt#27 has been on my 'shame list' for a while 😅 

Nice to see the `SignalGroup` out in the wild, really nicely separates event logic from the list itself!

PR #65, opened 2022-02-11
add a magicgui test to CI

tlambert03:
run magicgui tests during CI.  Though magicgui will no longer be trying to subclass psygnal anymore after magicgui 0.4.0, this will just make sure that any psygnal changes don't break magicgui (such as the _normalize_slot Attribute error in magicgui < 0.3.6 with psygnal >= 0.3.0


tlambert03:
benchmark failure is unrelated.  will merge.

PR #79, opened 2022-03-29
port `EventedDict` from napari

alisterburt:
This is pretty much a straight port of the `EventedDict` @lukasz-migas added to napari, using `SignalGroup` for the `DictEvents` rather than the napari event system

I removed a few little pieces that I wasn't sure about, namely
- child events
- `key` method on the dict
- the `update` event (changing/changed seem to cover the same cases)

Spent some time trying to please mypy too - typing is maybe starting to make some sense? 😆 I have likely made mistakes, please correct me if anything jumps out

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,7 +21,7 @@ repos:
       - id: autoflake
         args: ["--in-place", "--remove-all-unused-imports"]
   - repo: https://github.com/asottile/pyupgrade
-    rev: v2.31.0
+    rev: v2.31.1
     hooks:
       - id: pyupgrade
         args: [--py37-plus, --keep-runtime-typing]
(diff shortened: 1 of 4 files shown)


tlambert03 (approved):
Lovely job here too!  Thanks @alisterburt  (and @lukasz-migas for the prior work in napari).

I made a minor change to the `__init__`, to match the signature of `dict` (which, in addition to a mapping, accepts an iterable of (key, value) 2-tuples, as well as kwargs (like `dict(a=1, b=2)`)

Czaki:
> the `update` event (changing/changed seem to cover the same cases)

it is not covered. The evented dict pass events from value objects. And such object could have updated event. But as I see this implementation does not contain the propagation of events from dict values.

tlambert03:
Could go in a follow up if a failing test is found.  Thanks

PR #83, opened 2022-05-05
Evented pydantic model

### LINKED ISSUE #80: Is a pydantic `EventedModel` in scope
Would you consider a PR with a port of the `EventedModel` from napari?

In my mind the pydantic dependency would be gated behind an `extras_require` entry in setup called `pydantic`

tlambert03:
closes #80 

Adds an EventedModel based on the one in `napari`, using psygnal instead


tlambert03:
this is about ready if you'd want to take a glance @alisterburt.

Main differences between this and the napari version:
- `model.events` is now a `SignalGroup` instead of an `EmitterGroup` (which also means that the signals will emit the values directly, rather than an event object that has a `value` attribute)
- all of the funny `property` support stuff is there, but must be enabled with `allow_property_setters = True` in the Config.
- events for properties must be manually opted-into by either setting `guess_property_dependencies = True`, or by explicitly providing a `property_dependencies =` dict in the Config (as we do in napari in some places)

alisterburt:
Woah, had a quick look and this is stunning! Dealing with the pyside shenanigans can't have been fun.

Haven't given a proper review because packing for the US but super duper excited for this - see you soon!!

psygnal/_evented_model.py
```diff
@@ -0,0 +1,456 @@
...
+class EventedModel(BaseModel, metaclass=EventedMetaclass):
+    """A pydantic BaseModel that emits a signal whenever a field value is changed.
+
+    In addition to standard pydantic `BaseModel` properties (see
+    https://pydantic-docs.helpmanual.io/usage/models/), this class adds the following:
+
+    1. gains an `events` attribute that is an instance of `psygnal.SignalGroup`. This
+       group will have a signal for each field in the model (excluding private
+       attributes and non-mutable fields).  Whenever a field in the model is mutated,
+       the corresponding signal will emit with the new value (see example below).
```
alisterburt, line 214: we have objects in napari which provide multiple events (`changing`, `changed`) per 'evented' attribute -> is there some reason we don't need these now? I'm wondering if anyone will feel limited by the event only happening post value change (computing deltas might be hard?)

Definitely not suggesting this is necessary right now, just curious

tlambert03: The only reason I did that (in the evented list) was because qt needs to know both before and after that something is changing... mostly so that it can call [`beginMoveRows`](https://doc.qt.io/qt-5/qabstractitemmodel.html#beginMoveRows) beforehand, and [`endMoveRows`](https://doc.qt.io/qt-5/qabstractitemmodel.html#endMoveRows) at the end.   I don't particularly love having all those events otherwise :) so i guess I'd wait until something else needs that before implementing globally

psygnal/_signal.py
```diff
@@ -852,6 +854,12 @@ def paused(
         finally:
             self.resume(reducer, initial)
 
+    def __getstate__(self) -> dict:
+        """Return dict of current state, for pickle."""
+        d = {slot: getattr(self, slot) for slot in self.__slots__}
+        d.pop("_lock", None)
+        return d
```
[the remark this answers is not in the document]
tlambert03: perhaps!  But since that is _the_ mechanism for providing pickle support (https://docs.python.org/3/library/pickle.html#object.__getstate__) we'd definitely want to make sure that the dict returned by a napari layer was _actually_ fully serializable. (not just "almost" serializable)

PR #108, opened 2022-07-11
feat: add connect_setitem

tlambert03:
found myself needing this today.  @czaki, what do you think?


Czaki:
I like to add this to psygnal, but I have mixed feelings about setting multiple values by default. I prefer to at least change `maxargs` default value to 1, to force a user to explicitly allow more than one value to be set to a parameter.

Czaki:
Hmm. This may break existing code. 

Maybe better left it as it is. And for keep uniform interface do not change this in this PR.

PR #115, opened 2022-07-25
feat: raise exceptions as EmitLoopError

### LINKED ISSUE #106: Add option to continue past failures in connected callbacks
* psygnal version: 0.3.5
* Python version: 3.10.5
* Operating System: Linux

### Description
Largely in pursuit of a more robust MDAEngine (https://github.com/pymmcore-plus/pymmcore-plus/issues/130) I'd like to have the option to enable continuing with further signal emissions even after an exception in a callback. So the following should hopefully print `hi`

```python
from psygnal import Signal

class MyObject:
    value_changed = Signal(str)

my_obj = MyObject()

@my_obj.value_changed.connect
def bad_callback(new_value: str):
    raise ValueError("oh no")

@my_obj.value_changed.connect
def on_change(new_value: str):
    print(f"The value changed to {new_value}!")

my_obj.value_changed.emit('hi')
```

tlambert03:
closes #106 by raising all emit loop errors as a special `EmitLoopError` exception (RuntimeError) type:

@ianhi, this gives the decision on how to handle (or suppress) to the thing doing the emitting:

```python
from contextlib import suppress
from psygnal import EmitLoopError

with suppress(EmitLoopError):
    obj.emitter.emit(...)
```

sound good?

### DIFF
--- a/src/psygnal/__init__.py
+++ b/src/psygnal/__init__.py
@@ -16,6 +16,7 @@ __all__ = [
     "__version__",
     "_compiled",
     "EmissionInfo",
+    "EmitLoopError",
     "EventedModel",
     "Signal",
     "SignalGroup",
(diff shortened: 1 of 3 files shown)


tlambert03:
i feel pretty confident that this is a good way to go ... so merging

PR #118, opened 2022-08-02
feat: Add ability to disconnect slots from Signal group directly

alisterburt:
This PR adds a `SignalGroup.disconnect()` method which behaves as `Signal.disconnect()` does but for all signals in the group.

I'm circling back to the Qt stuff we talked about a while ago and this would be super convenient in some cases there :-)

I realise this method is overloading `SignalInstance.disconnect()` so the behaviour may be unexpected... open to alternative names

### DIFF
--- a/src/psygnal/_group.py
+++ b/src/psygnal/_group.py
@@ -10,7 +10,7 @@ the args that were emitted.
 from contextlib import contextmanager
 from typing import Any, Callable, Dict, Iterable, Iterator, Optional, Tuple, Union
 
-from psygnal._signal import Signal, SignalInstance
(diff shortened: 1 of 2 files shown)


tlambert03:
ok, cleaned the other stuff up.  Tests look good, but this unfortunately causes another (strong reference) fail.  I think that adding `super().disconnect(slot, missing_ok)` to the end of the `disconnect` method (after the for loop) will fix it.  But I admit, it wasn't immediately obvious to me why! :joy:

alisterburt:
hahaha love it, thanks for taking a look and tidying up! CI fail looks like it just needs to be re-run then this is good to go :)

PR #119, opened 2022-08-02
fix: fix listevents docstring parameter mismatch

alisterburt:


src/psygnal/containers/_evented_list.py
```diff
@@ -170,7 +170,7 @@ def __setitem__(self, key: Index, value: Union[_T, Iterable[_T]]) -> None:
                 raise TypeError("Can only assign an iterable to slice")
             value = [self._pre_insert(v) for v in value]  # before we mutate the list
         else:
-            value = self._pre_insert(cast(_T, value))
+            value = self._pre_insert(cast("_T", value))
```
alisterburt, line 173: pre-commit was complaining about this

tlambert03:
oh my... looks like we've got some other issues too :/  will take a look.  thanks!

alisterburt:
Sorry to bring problems to the forefront!

src/psygnal/containers/_evented_list.py
```diff
@@ -58,8 +58,8 @@ class ListEvents(SignalGroup):
         `(index, value)` emitted after `value` is removed at `index`
     moving : Signal[int, int]
         `(index, new_index)` emitted before an item is moved from `index` to `new_index`
-    moved : Signal[int, int]
-        `(index, new_index, value)` emitted after `value` is moved from `index` to
+    moved : Signal[tuple, int]
```
tlambert03, line 61: actually it looks like it should be `Signal[int, int, int]` ... see line 283, where the actual emit happens

PR #134, opened 2022-10-26
fix: fix disconnection of partials

### LINKED ISSUE #37: partial disconnection
* psygnal version: current development
* Python version:
* Operating System:

### Description

This test:
```python
@pytest.mark.parametrize(
    "slot",
    [
        "f_no_arg",
        "f_int_decorated_stupid",
        "f_int_decorated_good",
        "f_any_assigned",
        "partial",
    ],
)
def test_weakref_diconnect(slot):
    """Test that a connected method doesn't hold strong ref."""
    emitter = Emitter()
    obj = MyObj()

    assert len(emitter.one_int) == 0
    emitter.one_int.connect(
        partial(obj.f_int_int, 1) if slot == "partial" else getattr(obj, slot)
    )
    assert len(emitter.one_int) == 1
    emitter.one_int.emit(1)
    assert len(emitter.one_int) == 1
    emitter.one_int.disconnect(
        partial(obj.f_int_int, 1) if slot == "partial" else getattr(obj, slot)
    )
    assert len(emitter.one_int) == 0
``` 
does not pass for a `partial` parameter. 
There is a question if disconnection of a `partial` should be supported?

Posisble solutions (base on discussion):

1. add `until` parameter with signal triggering disconnection
```python
emitter.changed.connect(partial(self.something, 1), until=self.destroyed)
``` 	
2. add `disconnect_object` method which allow to disconnect all method of given object. 

First solution is nice and more pythonic for me, but introduce risk of introduce reference loop which will keep obsolete objects alive (or hit PyQt5 of C++ object was already deleted). Second one is simpler but increase disconnection code.

tlambert03:
fixes #37 ... sorry for the long delay on this one @Czaki!

note that I made a slight adjustment to your test in that issue (that makes it a little less awesome, but still better):

```python
    assert len(emitter.one_int) == 0
    # this is the change... storing the partial as `cb`
    # we have to connect/disconnect the *same* partial
    # which I think is a reasonable compromise... 
    cb = partial(obj.f_int_int, 1) if slot == "partial" else getattr(obj, slot)
    emitter.one_int.connect(cb)
    assert len(emitter.one_int) == 1
    emitter.one_int.emit(1)
    assert len(emitter.one_int) == 1
    emitter.one_int.disconnect(cb)
```

### DIFF
--- a/src/psygnal/_signal.py
+++ b/src/psygnal/_signal.py
@@ -750,6 +750,10 @@ class SignalInstance:
             idx = self._slot_index(slot)
             if idx != -1:
                 self._slots.pop(idx)
+                if isinstance(slot, PartialMethod):
+                    _PARTIAL_CACHE.pop(id(slot), None)
+                elif isinstance(slot, tuple) and callable(slot[2]):
+                    _prune_partial_cache()
             elif not missing_ok:
                 raise ValueError(f"slot is not connected: {slot}")
 
@@ -1262,6 +1266,13 @@ def _partial_weakref(slot_partial: PartialMethod) -> Tuple[weakref.ref, str, Cal
 
         _PARTIAL_CACHE[_id] = (ref, name, wrap)
     return _PARTIAL_CACHE[_id]
+
+
+def _prune_partial_cache() -> None:
+    """Remove any partial methods whose object has been garbage collected."""
+    for key, (ref, *_) in list(_PARTIAL_CACHE.items()):
+        if ref() is None:
+            del _PARTIAL_CACHE[key]
 
 
 def _get_method_name(slot: MethodType) -> Tuple[weakref.ref, str]:
--- a/src/psygnal/_signal.py
+++ b/src/psygnal/_signal.py
@@ -1242,16 +1242,26 @@ def _is_subclass(left: Type[Any], right: type) -> bool:
     return issubclass(left, right)
 
 
+_PARTIAL_CACHE: Dict[int, Tuple[weakref.ref, str, Callable]] = {}
+
+
 def _partial_weakref(slot_partial: PartialMethod) -> Tuple[weakref.ref, str, Callable]:
     """For partial methods, make the weakref point to the wrapped object."""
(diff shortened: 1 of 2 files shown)


src/psygnal/_signal.py
```diff
@@ -1242,16 +1242,26 @@ def _is_subclass(left: Type[Any], right: type) -> bool:
     return issubclass(left, right)
 
 
+_PARTIAL_CACHE: Dict[int, Tuple[weakref.ref, str, Callable]] = {}
```
Czaki, line 1245: I have significant worries about memory leaking here. There should be at least some cleaning steps executed from time to time.

tlambert03: what are you thinking would take up much memory here?  the key is an integer, the value is a string of a weakref, string, and callable... are you worried about the args/kwargs closure on the callable?

tlambert03: I added a _prune_cache method that will get called either when a partial is disconnected, or when an emit loop finds a dead weakfref

Czaki: My main worries are connected with scenarios like what happens in napari EventedModels. There are multiple steps of creating objects and destroying them shortly after. 

If I understand correctly if user often creates a short living object that will be removed later and them will be used connection, then this dictionary may grows?

tlambert03: > If I understand correctly if user often creates a short living object that will be removed later and them will be used connection, then this dictionary may grows?

before my last commit, yes, there was the possibility for it to grow... but my main point above is that this dictionary is _miniscule_... it's using integers and strings as values for the most part... not the original object.

however, after my last commit https://github.com/tlambert03/psygnal/pull/134/commits/221ad11db92503e0e08f26ff1e5222fd64b4e6b7

when a slot is disconnected (either explicitly using `disconnect` or implicitly when the `_run_emit_loop` method encounters a dead weakref) then `_prune_partial_cache()` will be called that trims the dict

PR #143, opened 2022-11-23
refactor: don't use metaclass for signal group

tlambert03:
uses `__init_subclass__` rather than metaclass for `SignalGroup`


tlambert03:
magicgui failure is unrelated

PR #148, opened 2022-11-25
build: use mypyc instead of cython

tlambert03:
redoing #147.

This actually works pretty well now... benchmarks are quite similar to cython, with a much cleaner build process.
but i'm running into some difficulties with the containers... one of which is:
https://github.com/python/mypy/issues/14188

so this will probably sit for a bit.


tlambert03:
closing in favor of #149

PR #150, opened 2022-11-29
Don't unblock/resume within nested contexts

hanjinliu:
Signals are unblocked on exit of `blocked` context, which will cause problematic signal emission within nested contexts.

```python
from psygnal import Signal
class A:
    sig = Signal(int)
a = A()

a.sig.connect(print)

with a.sig.blocked():
    with a.sig.blocked():
        a.sig.emit(0)
    a.sig.emit(1)  # <----------- will be emitted
```

This PR fixes the problem above by checking if the signal is blocked first, and restore the original state on exit.

### DIFF
--- a/src/psygnal/_signal.py
+++ b/src/psygnal/_signal.py
@@ -969,11 +969,13 @@ class SignalInstance:
         ...     t.sig.emit(3)
         >>> # results in obj.signal.emit({1, 2, 3})
         """
+        was_paused = self._is_paused
         self.pause()
         try:
             yield
         finally:
-            self.resume(reducer, initial)
+            if not was_paused:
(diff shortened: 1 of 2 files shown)


tlambert03:
thanks @hanjinliu, good one.

looks like tests are failing, but it's not your fault.  You've hit on a funny/unexpected bug in the compiled version that somehow didn't surface in #149

I can reproduce locally, and will try to fix and push soon

tlambert03:
hey @hanjinliu  ... I think I have a fix for the tests here, but can't merge to your PR.  can you merge https://github.com/hanjinliu/psygnal/pull/1

tlambert03 (approved):
thanks @hanjinliu!

PR #160, opened 2022-12-28
ci: fix macos build

tlambert03:


tlambert03:
closing.  I think is now fixed upstream in hatchling

see also https://github.com/psf/black/pull/3511

PR #168, opened 2023-02-12
refactor: use weakmethod instead of _get_method_name

tlambert03:
This replaces some complicated internal logic for storing a weak reference to a bound method with [`weakref.WeakMethod`](https://docs.python.org/3/library/weakref.html#weakref.WeakMethod) ... it _seems_ like this is what should have been used all along?  @Czaki  ... would appreciate your thoughts if you have a moment to check

### DIFF
--- a/src/psygnal/_signal.py
+++ b/src/psygnal/_signal.py
@@ -1265,6 +1265,7 @@ class _BoundMethodCaller(SlotCaller):
 
     def __init__(self, slot: MethodType, max_args: int | None = None) -> None:
         self._method_ref = weakref.WeakMethod(slot)
+        self._obj_ref = weakref.ref(slot.__self__)  # needed for equality check
         self._max_args = max_args
 
     def __call__(self, args: tuple[object, ...]) -> bool:
@@ -1279,7 +1280,8 @@ class _BoundMethodCaller(SlotCaller):
     def __eq__(self, other: object) -> bool:
         return (
             isinstance(other, _BoundMethodCaller)
-            and self._method_ref == other._method_ref
+            and self._method_ref() == other._method_ref()
+            and self._obj_ref == other._obj_ref
         )
 
     def slot(self) -> MethodType:
@@ -1294,6 +1296,7 @@ class _PartialMethodCaller(SlotCaller):
 
     def __init__(self, slot: PartialMethod, max_args: int | None = None) -> None:
         self._method_ref = weakref.WeakMethod(slot.func)
+        self._obj_ref = weakref.ref(slot.func.__self__)  # needed for equality check
         self._max_args = max_args
         self._partial_args = slot.args
         self._partial_kwargs = slot.keywords
@@ -1313,7 +1316,8 @@ class _PartialMethodCaller(SlotCaller):
     def __eq__(self, other: object) -> bool:
         return (
             isinstance(other, _PartialMethodCaller)
-            and self._method_ref == other._method_ref
+            and self._method_ref() == other._method_ref()
+            and self._obj_ref == other._obj_ref
         )
 
     def slot(self) -> PartialMethod:
--- a/src/psygnal/_signal.py
+++ b/src/psygnal/_signal.py
@@ -680,6 +680,8 @@ class SignalInstance:
         """Get index of `slot` in `self._slots`.  Return -1 if not connected."""
         with self._lock:
             normed = _slot_caller(slot)
+            # NOTE:
+            # the == method here relies on the __eq__ method of each SlotCaller subclass
             return next((i for i, s in enumerate(self._slots) if s == normed), -1)
 
     def disconnect(self, slot: Callable | None = None, missing_ok: bool = True) -> None:
@@ -1268,7 +1270,7 @@ class _BoundMethodCaller(SlotCaller):
     def __call__(self, args: tuple[object, ...]) -> bool:
         method = self._method_ref()
         if method is None:
-            return True  # object has changed?
+            return True  # object has been deleted
         if self._max_args is not None:
             args = args[: self._max_args]
         method(*args)
@@ -1300,7 +1302,7 @@ class _PartialMethodCaller(SlotCaller):
     def __call__(self, args: tuple[object, ...]) -> bool:
         method = self._method_ref()
         if method is None:
-            return True
+            return True  # object has been deleted
 
         if self._max_args is not None:
             args = args[: self._max_args]
--- a/src/psygnal/_signal.py
+++ b/src/psygnal/_signal.py
@@ -1278,13 +1278,14 @@ class _BoundMethodCaller(SlotCaller):
         self._max_args = max_args
 
     def __call__(self, args: tuple[object, ...]) -> bool:
-        try:
-            method = self.slot()
-        except RuntimeError:
+        obj = self._obj_ref()
+        func = self._func_ref()
+        if obj is None or func is None:
             return True
+
         if self._max_args is not None:
             args = args[: self._max_args]
-        method(*args)
+        func(obj, *args)
         return False
 
     def __eq__(self, other: object) -> bool:
@@ -1295,6 +1296,11 @@ class _BoundMethodCaller(SlotCaller):
         )
 
     def _method(self) -> MethodType | None:
+        """Reconstruct the original method.
+
+        Note: this isn't used above in __call__ because it's a bit slower
+        """
+        # sourcery skip: assign-if-exp, reintroduce-else
         obj = self._obj_ref()
         func = self._func_ref()
         if obj is None or func is None:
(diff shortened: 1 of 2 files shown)


src/psygnal/_signal.py
```diff
@@ -1262,14 +1262,11 @@ class _BoundMethodCaller(SlotCaller):
...
         self._max_args = max_args
 
     def __call__(self, args: tuple[object, ...]) -> bool:
-        obj = self._ref()
-        if obj is None:
-            return True
-        method = getattr(obj, self._method_name, None)
+        method = self._method_ref()
         if method is None:
             return True  # object has changed?
```
Czaki, line 1271: Comment to "object has been deleted?"

src/psygnal/_signal.py
```diff
@@ -1280,49 +1277,47 @@ def __call__(self, args: tuple[object, ...]) -> bool:
     def __eq__(self, other: object) -> bool:
         return (
             isinstance(other, _BoundMethodCaller)
-            and self._ref == other._ref
-            and self._method_name == other._method_name
+            and self._method_ref == other._method_ref
```
Czaki, line 1280: As I check this comparison always returns false if the object is removed. I do not find this in the documentation, but I'm not sure if it was important in this context.

Czaki (approved):
Looks good to me. Two small remarks, but I'm not sure if they are important.

src/psygnal/_signal.py
tlambert03: thanks for having a look!  let me give that a little extra consideration.

tlambert03: interesting... yeah, that _is_ different from the equality of a `weakref.ref` to an object (that comparison is `True` both before and after object deletion).   annoying :joy:

tlambert03: I added some more tests, and fixed it (by also storing an object ref for the purpose of equality)... will merge once tests pass

src/psygnal/_signal.py
```diff
@@ -1280,49 +1280,50 @@ def __call__(self, args: tuple[object, ...]) -> bool:
     def __eq__(self, other: object) -> bool:
         return (
             isinstance(other, _BoundMethodCaller)
-            and self._ref == other._ref
-            and self._method_name == other._method_name
+            and self._method_ref() == other._method_ref()
+            and self._obj_ref == other._obj_ref
```
Czaki, line 1284: if `self._method_ref() == other._method_ref()` is evaluated to fail then `            and self._obj_ref == other._obj_ref` will not change anything.

Czaki: nice move :)

tests/test_psygnal.py
```diff
@@ -886,12 +878,98 @@ def test_weakref_disconnect(slot):
...
+
+def test_slot_caller_equality():
+    """Slot callers should be equal only if they represent the same bound-method."""
+
+    class T:
+        def x(self):
+            ...
+
+    t1 = T()
+    t2 = T()
```
Czaki, line 956:
```suggestion
    t1 = T()
    t2 = T()
    t1_ref = ref(t1) 
    t2_ref = ref(t2)
    
```

tests/test_psygnal.py
```diff
@@ -886,12 +878,98 @@ def test_weakref_disconnect(slot):
...
+    def _assert_equality():
+        assert bmt1_a == bmt1_b
+        assert bmt2_a == bmt2_b
+        assert bmt1_a != bmt2_a
+        assert bmt1_b != bmt2_b
+
+    _assert_equality()
+    del t1
+    gc.collect()
+    _assert_equality()
```
Czaki, line 972:
```suggestion
    gc.collect()
    assert t1_ref() is None
    _assert_equality()
```

tests/test_psygnal.py
```diff
@@ -886,12 +878,98 @@ def test_weakref_disconnect(slot):
...
+    del t2
+    gc.collect()
```
Czaki, line 974:
```suggestion
    gc.collect()
    assert t2_ref() is None
```

Czaki (commented):
Maybe validate if objects are properly deleted?

tlambert03:
ooof... benchmarks have taken a serious hit. when emitting to methods.

```
before           after         ratio
[157](https://github.com/pyapp-kit/psygnal/actions/runs/4157067007/jobs/7191284976#step:5:158)
     [1fa88adb]       [cc7499d5]
[158](https://github.com/pyapp-kit/psygnal/actions/runs/4157067007/jobs/7191284976#step:5:159)
     <v0.7.2^0>                 
[159](https://github.com/pyapp-kit/psygnal/actions/runs/4157067007/jobs/7191284976#step:5:160)
+      32.3±0.1μs       62.6±0.9μs     1.94  benchmarks.EmitSuite.time_emit_to_method(70)
[160](https://github.com/pyapp-kit/psygnal/actions/runs/4157067007/jobs/7191284976#step:5:161)
+      39.3±0.5μs         71.9±1μs     1.83  benchmarks.EmitSuite.time_emit_to_partial(70)
[161](https://github.com/pyapp-kit/psygnal/actions/runs/4157067007/jobs/7191284976#step:5:162)
+     6.85±0.05μs       11.4±0.3μs     1.66  benchmarks.EmitSuite.time_emit_to_method(10)
[162](https://github.com/pyapp-kit/psygnal/actions/runs/4157067007/jobs/7191284976#step:5:163)
+     7.92±0.02μs       12.4±0.4μs     1.57  benchmarks.EmitSuite.time_emit_to_partial(10)
[163](https://github.com/pyapp-kit/psygnal/actions/runs/4157067007/jobs/7191284976#step:5:164)
+      57.8±0.2μs         89.6±1μs     1.55  benchmarks.EmitSuite.time_emit_to_all(70)
[164](https://github.com/pyapp-kit/psygnal/actions/runs/4157067007/jobs/7191284976#step:5:165)
+     10.5±0.03μs       15.0±0.2μs     1.43  benchmarks.EmitSuite.time_emit_to_all(10)
[165](https://github.com/pyapp-kit/psygnal/actions/runs/4157067007/jobs/7191284976#step:5:166)
+     2.86±0.01μs      3.45±0.04μs     1.21  benchmarks.EmitSuite.time_emit_to_method(1)
[166](https://github.com/pyapp-kit/psygnal/actions/runs/4157067007/jobs/7191284976#step:5:167)
+     2.97±0.01μs      3.57±0.02μs     1.20  benchmarks.EmitSuite.time_emit_to_partial(1)
[167](https://github.com/pyapp-kit/psygnal/actions/runs/4157067007/jobs/7191284976#step:5:168)
+     3.32±0.01μs      3.85±0.03μs     1.16  benchmarks.EmitSuite.time_emit_to_all(1)
```

if I can't fix that, I'll probably just stick with the original method

Czaki:
I cannot add suggestions for the import of `weakref`.

tlambert03:
re benchmarks: it looks like the `WeakMethod.__call__` actually _rebinds_ the function to the instance, rather than just getting the name of the function on the instance as we were doing before:

```python
In [9]: class T:
   ...:     def x(self):
   ...:         ...
   ...:

In [10]: t = T()

In [11]: r = weakref.ref(t)

In [12]: m = weakref.WeakMethod(t.x)

In [13]: %timeit getattr(r(), 'x')
93.9 ns ± 0.719 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [14]: %timeit m()
399 ns ± 12.1 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

# the weakref call method is kinda like this:
In [21]: fref = weakref.ref(T.x)

In [22]: %timeit MethodType(fref(), r())
158 ns ± 1.85 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

@Czaki  what do you think?  is is worth a small hit in performance to use a cleaner approach?

(I just pushed a slight change that uses the same general approach of `WeakMethod`... but re-implements it, to see if we can speed it up slightly with mypyc)

tlambert03:
oh, actually that helped quite a bit!

```python
before           after         ratio
     [1fa88adb]       [a4caa8bd]
     <v0.7.2^0>                 
+        38.1±2μs         46.2±3μs     1.21  benchmarks.EmitSuite.time_emit_to_method(70)
```

tlambert03:
last benchmark is at 18% slower in only one test: when binding a lot (70) of methods.  I'm inclined to go with this as it feels more "correct".

src/psygnal/_signal.py
```diff
@@ -1280,49 +1290,50 @@ def __call__(self, args: tuple[object, ...]) -> bool:
...
+            and self._func_ref == other._func_ref
         )
 
+    def _method(self) -> MethodType | None:
+        obj = self._obj_ref()
+        func = self._func_ref()
+        if obj is None or func is None:
+            return None
+        return self._method_type(func, obj)
```
Czaki, line 1302: Why need to rebind instead of directly calling `func(obj, *args)`? (requires bigger refactor of course)

tlambert03: ah, good idea.  lemme try that as well.  (`weakref.WeakMethod` obviously rebinds to return the "original" object... but you're right that we don't need that here)

tlambert03: that helped!  no benchmark regression 🎉

tlambert03: sidenote: I recently got my beta invite to https://codspeed.io ... will add here soon

tlambert03:
thanks a lot for the eyes @Czaki ... your input was very helpful.  will merge later today

PR #171, opened 2023-02-15
feat: weak_partial

tlambert03:
playing around a bit here.  this replace all of the various SlotCaller subtypes with a single `weak_partial` object that I can imagine wanting to use in other libraries (might make it a public export eventually).

### DIFF
--- a/asv.conf.json
+++ b/asv.conf.json
@@ -15,7 +15,7 @@
   ],
   "matrix": {
     "env": {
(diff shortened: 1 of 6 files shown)


tlambert03:
ha ... turns it it's very hard to touch anything here without decreasing performance :joy:

tlambert03:
see #178 instead

PR #188, opened 2023-02-22
refactor: change patching of __setattr__ in SignalGroupDescriptor, make more explicit

tlambert03:
adding `SignalGroupDescriptor` as a class attribute will patch the class's `__setattr__` method (on the first access of the descriptor on an instance).  This PR:

1. changes the moment when that patching occurs
2. provides an opt-out mechanism (in which case the user must patch `__settattr__` themselves)
3. documents it more obviously
4. uses a cleaner context manager within the `evented_setattr` pattern


tlambert03:
(magicgui test failure is only related to the bad version naming on the rc v0.8.0rc0 ... will fix in the next rc)

Issue #162, opened 2023-01-16
Problem with mypyc and PyInstaller


kasasxav:
* psygnal version: the one in napari
* Python version: 3.9
* Operating System: windows

### Description

I am trying to generate an executable file and psygnal is a dependency of it (through napari). It was working before but now I run this error when rying to run the executable:

"ModuleNotFoundError: No module named 'fa33d3e2c63a4523d3c5__mypyc"

I have looked into the code and it seems to be coming from the decorator of mypy in _evented_decorator.py import _group.py (https://github.com/pyapp-kit/psygnal/blob/df1f6f6397957a682634c59679f5a62a79b392c0/src/psygnal/_group.py#L43). 

I saw in [014a285056c5b86d22715cdfd7cc2274466d94c2 ](https://github.com/pyapp-kit/psygnal/commit/014a285056c5b86d22715cdfd7cc2274466d94c2) that you changed from cython to mypyc and I think this might explain why it was working before and not anymore.

### What I Did

```
pyinstaller --specpath . setup.py (in ImSwitch)

It compiles into the distribution but when double clicking it runs that error.
```

tlambert03:
yeah, @Czaki also had difficulties running pyinstaller with mypyc's approach.  he might have insight here

tlambert03:
(btw, I'll note that all of those compiled files are completely optional, so as a workaround, simply delete all the compiled (`.so`) files from the psygnal directory/subdirectories and you should be fine)

kasasxav:
Thanks for the quick answer! Deleting files sounds like something easy and interesting, in case they are optional. I don't manage to find them though, do you mean in site-packages? or in the pyinstaller dist folder? (in neither I found :l)

tlambert03:
> do you mean in site-packages? or in the pyinstaller dist folder? (in neither I found :l)

haven't used pyinstaller in a while, so I can't quite remember, but basically, the point is that you want to delete any compiled files (`.so` on linux, and not sure on windows) from within the psygnal directory, wherever pyinstaller would be looking to import it.  So, I'd assume that's the site-packages that pyinstaller imports from when it creates the frozen package.  (i think the dist folder would be too late).

btw, deleting those compiled files would have the same effect as installing with `pip install psygnal --no-binary psygnal` in the first place

kasasxav:
I deleted that but it fails because there is an import from a button within magicgui that needs to import psygnal.
I tried commenting the decorator inside the _group class but then I run with this error:

[...]

PR #194, opened 2023-02-27
build: add pyinstaller hook to simplify frozing apps using pyinstaller 

Czaki:
see https://github.com/pyapp-kit/psygnal/issues/162#issuecomment-1446344343

I have created code to automatically add every `.so`, `pyd` file to hidden imports. 

It is a little complicated but reduces maintenance. Maybe I should just hardcode the list?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -168,7 +168,7 @@ testpaths = ["tests"]
 filterwarnings = [
     "error",
     "ignore:The distutils package is deprecated:DeprecationWarning:",
-    "ignore:BackendFinder.find_spec() not found; falling back to find_module()", # pyinstaller import
+    "ignore:BackendFinder.find_spec()", # pyinstaller import
 ]
 
 # https://mypy.readthedocs.io/en/stable/config_file.html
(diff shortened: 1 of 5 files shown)


tlambert03:
> It is a little complicated but reduces maintenance. Maybe I should just hardcode the list?

I'm happy to go with whatever you suggest in this case.  I don't mind the complicated file (as long as it's ok if I ping you in the future if i have questions 😄 )

Since psygnal has a strict 100% test coverage check, it's also ok with me if you want to ignore `hook_psygnal` from coverage

Czaki:
> Since psygnal has a strict 100% test coverage check, it's also ok with me if you want to ignore `hook_psygnal` from coverage

The stage part here is that `hook-pysgnal` is at least partially covered (by `test_hook_content`), but coverage fails to discover it.  But I cannot use normal import because of `-` in the name required by pyinstaller.

tlambert03:
Ha! Interesting edge case 🫠

tlambert03 (approved):

Issue #198, opened 2023-03-13
Feature: connection across threads


tlambert03:
In some cases, it is desirable to use signals to communicate between objects across threads.
Currently, psygnal works fine for emitting to receivers within the same thread, but when an emitter exists in a subthread and wants to signal back to the main thread, tricks are needed.  It would be nice to implement a convenience around this 

Qt supports these signal-slot connection types:

- [Auto Connection](https://doc.qt.io/qt-6/qt.html#ConnectionType-enum) (default) If the signal is emitted in the thread which the receiving object has affinity then the behavior is the same as the [Direct Connection](https://doc.qt.io/qt-6/qt.html#ConnectionType-enum). Otherwise, the behavior is the same as the [Queued Connection](https://doc.qt.io/qt-6/qt.html#ConnectionType-enum)."
- Direct Connection The slot is invoked immediately, when the signal is emitted. The slot is executed in the emitter's thread, which is not necessarily the receiver's thread.
- Queued Connection The slot is invoked when control returns to the event loop of the receiver's thread. The slot is executed in the receiver's thread.
- [Blocking Queued Connection](https://doc.qt.io/qt-6/qt.html#ConnectionType-enum) The slot is invoked as for the Queued Connection, except the current thread blocks until the slot returns.

it would be nice to implement similar stuff here.  Currently we have "direct connection" ... but would be nice to have queued connection.  It's possible that will be significantly harder here without the whole `QObject` parent object

tlambert03:
with #200, I'm happy with the current state of connecting across threads, so will close this.  We don't yet have "blocking queued connections" but I'm not sure it's an important feature.

PR #199, opened 2023-04-02
feat: add `queue` flag to `SignalInstance.emit` method

tlambert03:
This PR attempts to provide a partial solution<sup>*</sup> to the problem of needing to have an emitter run in another thread (e.g. imagine a thread that polls some device and emits change events), while having the callbacks called in some other thread (like the main GUI thread in Qt for example).  One could certainly use [superqt thread workers](https://pyapp-kit.github.io/superqt/utilities/threading/?h=thread#superqt.utils.thread_worker), but this provides a slightly lower level (and possibly more performant?) solution.

signal instances now have a private `Queue` object, and the `emit` method gains a new `queue=...` argument, which is False by default.  When True, the emission does not occur immediately, but rather is added to the queue to be emitted at later when something calls `emit_queued` (currently, it is up to the user to call that in whatever thread they want).  So, an example using a QTimer to emit queued arguments in the main thread would be:

```python
from qtpy.QtCore import QTimer, Qt
from psygnal import Signal, emit_queued
import threading

class Emitter:
    sig = Signal(int)

obj = Emitter()

@obj.sig.connect
def on_emit(value: int):
    print(f"got value {value} from thread {threading.current_thread().name!r}")

# this timer lives in the main thread
timer = QTimer()

# whenever the timer times out, it will call `emit_queued`
timer.timeout.connect(emit_queued)
timer.start(0)  # start the timer

def _some_thread_that_emits() -> None:
    # inside the thread call `emit()` with `queue=True`
    obj.sig.emit(1, queue=True)

threading.Thread(target=_some_thread_that_emits).start()
```

@kephale, hopefully this will be useful for pulser.  @Czaki, if you have a moment to take a quick look, I'd appreciate your insights

----------- 

<sup>*</sup> I say "partial" solution because it's still up to the user to invoke the emission from the main thread.  So we haven't quite provided a complete solution to #198 ... but then again, I'm not sure we can do that entirely in psygnal.  Because psygnal itself doesn't manage an event loop, we can likely just "play nicely" with various event loop patterns

### DIFF
--- a/src/psygnal/_signal.py
+++ b/src/psygnal/_signal.py
@@ -7,6 +7,7 @@ import weakref
 from contextlib import contextmanager, suppress
 from functools import lru_cache, reduce
 from inspect import Parameter, Signature, isclass
+from queue import Queue
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -330,6 +331,7 @@ class SignalInstance:
         self._is_blocked: bool = False
         self._is_paused: bool = False
         self._lock = threading.RLock()
+        self._emit_queue: Queue[tuple[tuple, dict]] = Queue()
 
     @property
     def signature(self) -> Signature:
@@ -815,9 +817,10 @@ class SignalInstance:
     def emit(
         self,
         *args: Any,
-        check_nargs: bool = False,
-        check_types: bool = False,
-        asynchronous: Literal[False] = False,
+        queue: bool = ...,
+        check_nargs: bool = ...,
+        check_types: bool = ...,
+        asynchronous: Literal[False] = ...,
     ) -> None:
         ...  # pragma: no cover
 
@@ -825,8 +828,9 @@ class SignalInstance:
     def emit(
         self,
         *args: Any,
-        check_nargs: bool = False,
-        check_types: bool = False,
+        queue: bool = ...,
+        check_nargs: bool = ...,
+        check_types: bool = ...,
         asynchronous: Literal[True],
     ) -> EmitThread | None:
         # will return `None` if emitter is blocked
@@ -835,6 +839,7 @@ class SignalInstance:
     def emit(
         self,
         *args: Any,
+        queue: bool = False,
         check_nargs: bool = False,
         check_types: bool = False,
         asynchronous: bool = False,
@@ -850,6 +855,12 @@ class SignalInstance:
         *args : Any
             These arguments will be passed when calling each slot (unless the slot
             accepts fewer arguments, in which case extra args will be discarded.)
+        queue: bool
+            If `False` (default), invoke connected callbacks immediately. If `True`,
+            add the arguments to a queue to be emitted the next time `emit_queued` is
+            called. This is useful when emitting from a thread other than the main
+            thread.  It is up to the user to call `emit_queued` at the appropriate time,
+            (psygnal doesn't do it for you).
         check_nargs : Optional[bool]
             If `False` and the provided arguments cannot be successfully bound to the
             signature of this Signal, raise `TypeError`.  Incurs some overhead.
@@ -897,6 +908,13 @@ class SignalInstance:
 
             SignalInstance._debug_hook(EmissionInfo(self, args))
 
+        if queue:
+            with self._lock:
+                self._emit_queue.put(
+                    (args, {"queue": False, "asynchronous": asynchronous})
+                )
+            return None
+
         if asynchronous:
             sd = EmitThread(self, args)
             sd.start()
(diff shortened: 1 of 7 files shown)


src/psygnal/_signal.py
```diff
@@ -897,6 +908,13 @@ def emit(
 
             SignalInstance._debug_hook(EmissionInfo(self, args))
 
+        if queue:
+            with self._lock:
+                self._emit_queue.put(
+                    (args, {"queue": False, "asynchronous": asynchronous})
+                )
```
Czaki, line 915: Python Queues are thread safe (they are designed among others to thread communication). So using an additional thread looks obsolete.

Czaki (commented):
Hm. Why did you use a private queue instead of a global one? It looks like Qt uses one queue per thread.

src/psygnal/_signal.py
tlambert03: ok! thanks!

tlambert03:
> Why did you use a private queue instead of a global one? It looks like Qt uses one queue per thread.

i didn't want to make it Qt specific... but maybe I can offer a solution that gives the best of both worlds?  For example, maybe if `emit(queue: bool | QueueLike = False)` ... if `queue == True` then the internal queue is used, otherwise whatever queue is provided is used?  Can you give me a tip as to where one would access that one global queue in Qt?

tlambert03:
hmm... getting pyinstaller fails too

Czaki:
> Can you give me a tip as to where one would access that one global queue in Qt?

I am investigating it currently, but using a global queue may allow using a single timer for all objects that should emit a signal in this thread, as the current solution requires micromanagement on timers.

Czaki:
> hmm... getting pyinstaller fails too

As pyinstaller cannot inspect `_signal` as it is compiled then it may not include `queue` module. Similar thing to current adding `__future__` to hiddenimports https://github.com/pyapp-kit/psygnal/blob/74aa34f94529ab2414428ab7148a73536aafeff1/src/psygnal/_pyinstaller_util/hook-psygnal.py#L27

tlambert03:
> > hmm... getting pyinstaller fails too
> 
> As pyinstaller cannot inspect `_signal` as it is compiled then it may not include `queue` module. Similar thing to current adding `__future__` to hiddenimports
> 
> https://github.com/pyapp-kit/psygnal/blob/74aa34f94529ab2414428ab7148a73536aafeff1/src/psygnal/_pyinstaller_util/hook-psygnal.py#L27

thanks!  fixing

tlambert03:
> I am investigating it currently, but using a global queue may allow using a single timer for all objects that should emit a signal in this thread, as the current solution requires micromanagement on timers.

let me know if you learn anything @Czaki  ... I'm not sure I was following when you said "It looks like Qt uses one queue per thread.", but I can definitely see the merits of having a single global psygnal queue object, and a global `emit_queued` function that could be connected to a single Timer over what I currently have here.

i'll do that soon unless you have a better qt-specific suggestion

Czaki:
After rethink. It Qt, there is no signal queue, but the receiver queue (python functions or objects livings in the same thread is executed immediately. So maybe a better solution could be to implement this in connect, instead of emit, so an actress could decide about mode when connecting?

https://doc.qt.io/qt-5/threads-qobject.html#signals-and-slots-across-threads

tlambert03:
I do like that generally idea.  but i think the real problem there is the fact that we don't have the concept of object thread affinity in psygnal.  Most of the logic of signals/slots across threads (queued and direct connections etc) is put on `QObject` itself  ... i.e. the thing that *has* the Signal ... not the signal itself.  So, without making more assumptions about the user class that contains a signal, I'm not sure we can do tricks like that so easily?  (maybe we could think of something clever... but I haven't come up with anything clean yet).

I do think a single global queue would definitely be better than nothing though (implementing now)

Czaki:
```python
class QueueConnector:
	def __init__(self):
		self.queue = Queue()
		
	def callback(self, callback, val):
        self.queue.put((callback, val))

    def connect(self, callback):
		#change callback to weakref if need
		return partial(self.callback, callback)

	def trigger_callbacks(self):
		while not self.queue.empty():
			callback, val = self.queue.get()
			# get original callback if need
			callback(val)

class Emitter:
    sig = Signal(int)

obj = Emitter()


def on_emit(value: int):
    print(f"got value {value} from thread {threading.current_thread().name!r}")

connector = QueueConnector()

obj.sig.connect(connector.connect(on_emit))

timer = QTimer()
timer.timeout.connect(connector.trigger_callbacks)
timer.start(0)  # start the timer


def _some_thread_that_emits() -> None:
    # inside the thread call `emit()` with `queue=True`
    obj.sig.emit(1, queue=True)

# start the thread 
threading.Thread(target=_some_thread_that_emits).start()
```

Could you let me know what you thought about this idea? 

This example does not support automatic disconnect on object deletion, but it may be done by raising a proper exception in `QueueConnector.callback`. 

(names of methods should be fixed)

tlambert03:
> Could you let me know what you thought about this idea?

i think i like it! playing with it now.  I can definitely see the benefits here of having the `connect` method be the one to determine the thread affinity.  I think I'd want to hide the `QueueConnector` object from the end-user, instead using an argument to `connect` (something like the connection type above)... is that your vision here as well?

thanks!

Czaki:
> r. I think I'd want to hide the `QueueConnector` object from the end-user, instead using an argument to `connect` (something like the connection type above)... is that your vision here as well?

The general problem with hide is that user is responsible for the integration of the queue with the event loop, as someone may want to use it with Qt, someone with async, someone with wxPython, etc. 

So there is a need to provide some integration interface.

Czaki:
But maybe the user may pass a queue as connect argument?

tlambert03:
> So there is a need to provide some integration interface.

right, I agree... that would be something like `psygnal.emit_queued()` (similar to the current PR).  That bit doesn't change,  this bit just changes exactly where and how we put stuff to the queue.  There will still be a need for each event loop to have something periodically clear that queue in the main thread (or wherever)

tlambert03:
> But maybe the user may pass a queue as connect argument?

yeah, I was thinking to have a `type` argument to the `connect` method, just like https://doc.qt.io/qt-6/qobject.html#connect

tlambert03:
will make an alternative PR for you to look at.  thanks again for your input here

Czaki:
> something periodically clear that queue in the main thread (or wherever)

So first, you prefer to avoid supporting the queue per thread on the first try?

tlambert03:
> So first, you prefer to avoid supporting the queue per thread on the first try?

umm, i don't think i do?  i probably wasn't clear above.  hang on, about to push something

tlambert03:
closing in favor of #200

PR #200, opened 2023-04-02
feat: add thread parameter to connection method, allowed "queued connections"

tlambert03:
alternative to #199  ... hat-tip to @Czaki  for the inspiration.

somewhat closer to realizing #198 

run this script as an example
```python
import threading
from qtpy.QtWidgets import QApplication
from psygnal import Signal
# special qt support, so everyone can share the same main-thread timer
from psygnal.qt import start_emitting_from_queue 


class Emitter:
    sig = Signal(int)

obj = Emitter()

@obj.sig.connect(thread='main')
def on_emit(value: int) -> None:
    """Callback function that states its thread."""
    print(f"got value {value} from thread {threading.current_thread().name!r}")

app = QApplication([])
start_emitting_from_queue()
threading.Thread(target=obj.sig.emit, args=(1,)).start()
app.processEvents()
```

### DIFF
--- a/src/psygnal/_queue.py
+++ b/src/psygnal/_queue.py
@@ -1,5 +1,5 @@
 from queue import Queue
-from typing import Any, Callable, Tuple
+from typing import Any, Callable, ClassVar, Tuple
 
 from ._weak_callback import WeakCallback
 
@@ -8,7 +8,7 @@ CbArgsTuple = Tuple[Callback, tuple]
 
 
 class QueuedCallback(WeakCallback):
-    _GLOBAL_QUEUE: Queue[CbArgsTuple] = Queue()
+    _GLOBAL_QUEUE: ClassVar[Queue[CbArgsTuple]] = Queue()
 
     def __init__(self, wrapped: WeakCallback) -> None:
         self._wrapped = wrapped
@@ -18,7 +18,7 @@ class QueuedCallback(WeakCallback):
         self._on_ref_error = wrapped._on_ref_error
 
     def cb(self, args: tuple = ()) -> None:
-        self._GLOBAL_QUEUE.put((self._wrapped.cb, args))
+        QueuedCallback._GLOBAL_QUEUE.put((self._wrapped.cb, args))
 
     def dereference(self) -> Callable | None:
         return self._wrapped.dereference()
--- a/src/psygnal/_queue.py
+++ b/src/psygnal/_queue.py
@@ -1,0 +1,35 @@
+from queue import Queue
+from typing import Any, Callable, Tuple
+
+from ._weak_callback import WeakCallback
+
+Callback = Callable[[tuple[Any, ...]], Any]
+CbArgsTuple = Tuple[Callback, tuple]
+
+
+class QueuedCallback(WeakCallback):
+    _GLOBAL_QUEUE: Queue[CbArgsTuple] = Queue()
+
+    def __init__(self, wrapped: WeakCallback) -> None:
+        self._wrapped = wrapped
+        self._key: str = wrapped._key
+        self._max_args: int | None = wrapped._max_args
+        self._alive: bool = wrapped._alive
+        self._on_ref_error = wrapped._on_ref_error
+
+    def cb(self, args: tuple = ()) -> None:
+        self._GLOBAL_QUEUE.put((self._wrapped.cb, args))
+
+    def dereference(self) -> Callable | None:
+        return self._wrapped.dereference()
+
+
+def emit_queued(queue: Queue[CbArgsTuple] = QueuedCallback._GLOBAL_QUEUE) -> None:
+    from ._signal import EmitLoopError
+
+    while not queue.empty():
+        cb, args = queue.get()
+        try:
+            cb(args)
+        except Exception as e:
+            raise EmitLoopError(slot_repr=repr(cb), args=args, exc=e) from e
--- a/src/psygnal/_queue.py
+++ b/src/psygnal/_queue.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from queue import Queue
+from threading import main_thread, Thread
 from typing import Any, Callable, ClassVar, Tuple
 
 from ._weak_callback import WeakCallback
@@ -10,9 +11,9 @@ CbArgsTuple = Tuple[Callback, tuple]
 
 
 class QueuedCallback(WeakCallback):
-    _GLOBAL_QUEUE: ClassVar[Queue[CbArgsTuple]] = Queue()
+    _GLOBAL_QUEUE: ClassVar[dict[Thread, Queue[CbArgsTuple]]] = defaultdict(Queue)
 
-    def __init__(self, wrapped: WeakCallback) -> None:
+    def __init__(self, wrapped: WeakCallback, thread: Thread = maint_thread()) -> None:
         self._wrapped = wrapped
         self._key: str = wrapped._key
         self._max_args: int | None = wrapped._max_args
@@ -20,15 +21,17 @@ class QueuedCallback(WeakCallback):
         self._on_ref_error = wrapped._on_ref_error
 
     def cb(self, args: tuple = ()) -> None:
-        QueuedCallback._GLOBAL_QUEUE.put((self._wrapped.cb, args))
+        QueuedCallback._GLOBAL_QUEUE[self._thread].put((self._wrapped.cb, args))
 
     def dereference(self) -> Callable | None:
         return self._wrapped.dereference()
 
 
-def emit_queued(queue: Queue[CbArgsTuple] = QueuedCallback._GLOBAL_QUEUE) -> None:
+def emit_queued(queue: Optional[Queue[CbArgsTuple]] = None) -> None:
     from ._signal import EmitLoopError
 
+if queue is None:
+    queue = QueuedCallback._GLOBAL_QUEUE[current_thread()]
     while not queue.empty():
         cb, args = queue.get()
         try:
--- a/src/psygnal/_queue.py
+++ b/src/psygnal/_queue.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from queue import Queue
-from threading import main_thread, Thread
+from threading import Thread, main_thread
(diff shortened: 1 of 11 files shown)


src/psygnal/_signal.py
```diff
@@ -356,6 +356,7 @@ def __repr__(self) -> str:
     def connect(
         self,
         *,
+        type: Literal["direct", "queued"] = "direct",
```
Czaki, line 359:
```suggestion
        type: Union[Literal["direct", "queued"], Thread] = "direct",
```

src/psygnal/_signal.py
```diff
@@ -369,6 +370,7 @@ def connect(
         self,
         slot: F,
         *,
+        type: Literal["direct", "queued"] = "direct",
```
Czaki, line 373:
```suggestion
        type: Union[Literal["direct", "queued"], Thread] = "direct",
```

src/psygnal/_signal.py
```diff
@@ -381,6 +383,7 @@ def connect(
         self,
         slot: F | None = None,
         *,
+        type: Literal["direct", "queued"] = "direct",
```
Czaki, line 386:
```suggestion
        type: Union[Literal["direct", "queued"], Thread] = "direct",
```

src/psygnal/_signal.py
```diff
@@ -415,6 +418,10 @@ def my_function():
         check_nargs : Optional[bool]
             If `True` and the provided `slot` requires more positional arguments than
             the signature of this Signal, raise `TypeError`. by default `True`.
+        type: {'direct', 'queued'}, optional
```
Czaki, line 421:
```suggestion
        type: {'direct', 'queued'} or Thread, optional
```

src/psygnal/_signal.py
```diff
@@ -415,6 +418,10 @@ def my_function():
...
+            If 'direct' (the default), this slot will be invoked immediately when a
+            signal is emitted.  If 'queued', invocation of this slot will be delayed
+            until the next time the event loop is entered.
```
Czaki, line 424:
```suggestion
            signal is emitted.  If 'queued', invocation of this slot will be delayed
            until the next time the main thread event loop is entered. If a thread 
            object is provided, then the event will be delayed to the next thread 
            event loop processing. 
```

src/psygnal/_signal.py
```diff
@@ -487,7 +494,12 @@ def _wrapper(
                     finalize=self._try_discard,
                     on_ref_error=_on_ref_err,
                 )
-                self._slots.append(cb)
+                if type == "queued":
+                    from ._queue import QueuedCallback
+
+                    self._slots.append(QueuedCallback(cb))
+                else:
+                    self._slots.append(cb)
```
Czaki, line 502:
```suggestion
                if type == "direct":
                    self._slots.append(cb)
                elif type == "queued":
                    from ._queue import QueuedCallback

                    self._slots.append(QueuedCallback(cb))
                else:
                    self._slots.append(QueuedCallback(cb, thread=type))
```

src/psygnal/qt.py
```diff
@@ -0,0 +1,42 @@
+from __future__ import annotations
```
Czaki, line 1:
```suggestion
from __future__ import annotations
from threading import current_thread, Thread
```

src/psygnal/qt.py
```diff
@@ -0,0 +1,42 @@
...
+
+try:
+    from qtpy.QtCore import Qt, QTimer
+except ImportError:
+    raise ImportError(
+        "The psygnal.qt module requires qtpy and some Qt backend to be installed"
+    ) from None
+
+_TIMER: QTimer | None = None
```
Czaki, line 10:
```suggestion
_TIMER: dict[Tread, QTimer] = {}
```

src/psygnal/qt.py
```diff
@@ -0,0 +1,42 @@
...
+        _TIMER = QTimer()
+        from ._queue import emit_queued
+
+        _TIMER.timeout.connect(emit_queued)
+
+    _TIMER.setTimerType(timer_type)
+    if _TIMER.isActive():
+        _TIMER.setInterval(msec)
+    else:
+        _TIMER.start(msec)
```
Czaki, line 35:
```suggestion
    thread = current_thread()
    if thread not in _TIMER:
        _TIMER[thread] = QTimer()
        from ._queue import emit_queued

        _TIMER[thread].timeout.connect(emit_queued)

    _TIMER[thread].setTimerType(timer_type)
    if _TIMER[thread].isActive():
        _TIMER[thread].setInterval(msec)
    else:
        _TIMER[thread].start(msec)
```

src/psygnal/qt.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def stop_emitting_from_queue() -> None:
+    """Stop the QTimer that monitors the global emission queue."""
+    global _TIMER
+    if _TIMER is not None:
+        _TIMER.stop()
```
Czaki, line 42:
```suggestion
    global _TIMER
    timer = _TIMER.get(current_thread())
    if timer is not None:
        timer.stop()
```

src/psygnal/_queue.py
```diff
@@ -0,0 +1,35 @@
...
+    def cb(self, args: tuple = ()) -> None:
+        QueuedCallback._GLOBAL_QUEUE.put((self._wrapped.cb, args))
+
+    def dereference(self) -> Callable | None:
+        return self._wrapped.dereference()
+
+
+def emit_queued(queue: Queue[CbArgsTuple] = QueuedCallback._GLOBAL_QUEUE) -> None:
+    from ._signal import EmitLoopError
+
```
Czaki, line 31:
```suggestion

if queue is None:
    queue = QueuedCallback._GLOBAL_QUEUE[current_thread()]
```

Czaki (commented):
I try to suggest an extension of this to provide a queue per thread. What did you think?

tlambert03:
Thanks! Will look at this closer tomorrow.  Do you generally like this pattern better on the whole?

Czaki:
Yes. I prefer it as the user decides if use queued connections.

src/psygnal/_queue.py
```diff
@@ -0,0 +1,41 @@
+from __future__ import annotations
+
+from queue import Queue
+from threading import Thread
```
Czaki, line 4:
```suggestion
from threading import main_thread, Thread
```

src/psygnal/_queue.py
```diff
@@ -0,0 +1,41 @@
...
+from ._weak_callback import WeakCallback
+
+Callback = Callable[[Tuple[Any, ...]], Any]
+CbArgsTuple = Tuple[Callback, tuple]
+
+
+class QueuedCallback(WeakCallback):
+    _GLOBAL_QUEUE: ClassVar[dict[Thread, Queue[CbArgsTuple]]] = defaultdict(Queue)
+
+    def __init__(self, wrapped: WeakCallback, thread: Thread = maint_thread()) -> None:
```
Czaki, line 16:
```suggestion
    def __init__(self, wrapped: WeakCallback, thread: Thread = main_thread()) -> None:
```

Czaki (commented):
there was bug in my comments

tlambert03:
question regarding API on the connect method @Czaki ...

I find I don't love the name `type` (other than the fact that Qt uses the name `type`).  One alternative would be to use 

```
def connect(self, ..., thread: Thread | None = None): ...
```

where `thread` is whatever thread you want this callback to be invoked in.
- `thread=None` would mean invoke this slot on whatever thread the signal was emitted from. 
- `threading=threading.main_thread()` or `threading=threading.current_thread()` does what it sounds like.

we could also accept the special strings `'main'` or `'current'` as conveniences to avoid needing to import from threading.

or do you prefer the words `"direct"` and `"queued"`?

Czaki:
I like this suggestion to pass thread as primary type.

tlambert03:
> I like this suggestion to pass thread as primary type.

so, *no* conveniences around string?  user *must* import from threading in order to use it?

Czaki:
Ys. Support for main and current is a good idea. As we do not have idea for moving object semantics theni it is better than direct and queued.

tlambert03:
one more question.  I'm inclined to default to the following behavior (most similar to "auto connection" on Qt):

- if `connect(callback, thread=some_thread)` is used, then when a signal is emitted on `some_thread`, then `callback` is invoked immediately (rather than being queued for later callback).

This avoids the surprising behavior of using `connect(callback, thread=main_thread())` in a single-threaded program, and then never seeing `callback` get called (unless you then specifically went in and used `emit_queued(main_thread())`). 

in code, the QueuedCallback object would then look like this:

```python
class QueuedCallback(WeakCallback):
    def cb(self, args: tuple = ()) -> None:
        if current_thread() is self._thread:
            self._wrapped.cb(args)
        else:
            QueuedCallback._GLOBAL_QUEUE[self._thread].put((self._wrapped.cb, args))

```

do you agree?

Czaki:
Hm. It looks like the best solution for now.

Czaki:
maybe also https://github.com/pyapp-kit/psygnal/blob/main/docs/usage.md should be updated?

src/psygnal/qt.py
```diff
@@ -0,0 +1,76 @@
...
+"""
+from __future__ import annotations
+
+from threading import Thread, current_thread
+
+from ._queue import emit_queued
+
+try:
+    from qtpy.QtCore import Qt, QTimer
+except ImportError:
```
Czaki, line 16: If I correctly remeber for part of `qtpy` (it looks like even `2.3.0`, when the latest is `2.3.1`) lack of Qt binding is signaled by `RuntimeError`, not `ImportError`

tlambert03:
> maybe also https://github.com/pyapp-kit/psygnal/blob/main/docs/usage.md should be updated?

i'll do that in a follow up.  I think i will also deprecate the `asynchronous` parameter on the emit method.  Someone can always just manually emit in a different thread.  The more important thing was always where the callback gets called and this solves that much better.

tlambert03:
excited about this!  thanks a lot for your input @Czaki

PR #225, opened 2023-08-12
test: python 3.12

tlambert03:
(good work numpy folks on dropping distutils!)


tlambert03:
working, except for compiled versions, which might have something to do with qt?

tlambert03:
https://github.com/python/mypy/issues/15277

PR #228, opened 2023-08-14
fix: fix signature inspection on debounced/throttled, update typing and wrapped

tlambert03:
in [mypyc's words](https://mypyc.readthedocs.io/en/latest/differences_from_python.html#introspection): "`inspect.signature` chokes on compiled functions"... which is a bummer for throttled/debounced if we're going to use them as callbacks (which is their main goal).

This PR fixes https://github.com/napari/napari/issues/6149#issuecomment-1678081559 by telling inspect.signature how to get to the wrapped function.  I think it should be good enough, but if more problems arise, we can just skip mypyc compilation on these modules.

cc @andy-sweet

*edit*: i just decided not to compile that module for now.

### DIFF
--- a/src/psygnal/_throttler.py
+++ b/src/psygnal/_throttler.py
@@ -32,7 +32,7 @@ class _ThrottlerBase:
 
         # this mimics what functools.wraps does, but avoids __dict__ usage and other
         # things that won't work with mypyc... HOWEVER, most of these dynamic
-        # assignments won't work in mypyc anyway
+        # assignments won't work in mypyc anyway (they just do nothing.)
         self.__module__: str = getattr(func, "__module__", "")
         self.__name__: str = getattr(func, "__name__", "")
         self.__qualname__: str = getattr(func, "__qualname__", "")
--- a/src/psygnal/_throttler.py
+++ b/src/psygnal/_throttler.py
@@ -1,9 +1,12 @@
 from __future__ import annotations
 
 from threading import Timer
-from typing import Any, Callable
+from typing import TYPE_CHECKING, Any, Callable
 
 from typing_extensions import Literal
+
+if TYPE_CHECKING:
+    import inspect
 
 Kind = Literal["throttler", "debouncer"]
 EmissionPolicy = Literal["trailing", "leading"]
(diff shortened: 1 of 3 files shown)


src/psygnal/_throttler.py
```diff
@@ -27,9 +30,18 @@ def __init__(
         self._args: tuple[Any, ...] = ()
         self._kwargs: dict[str, Any] = {}
 
+        # this mimics what functools.wraps does, but avoids __dict__ usage and other
+        # things that won't work with mypyc... HOWEVER, most of these dynamic
```
andy-sweet, line 34: By won't work, I guess you mean that mypyc will just ignore these? But won't error?

tests/test_throttler.py
```diff
@@ -79,3 +83,30 @@ def test_flush() -> None:
...
+    # make sure we can still inspect the signature
+    assert signature(f1).parameters["x"] == Parameter(
+        "x", Parameter.POSITIONAL_OR_KEYWORD, annotation=int
+    )
+
+    # make sure these are connectable
+    sig = SignalInstance((int, int, int))
+    sig.connect(f1)
+    sig.emit(1, 2, 3)
+    time.sleep(0.1)
```
andy-sweet, line 106: Always sad to see hopeful sleeps in tests, but I guess there's no equivalent of `pytest.waitSignal` in psygnal? Doesn't block this PR, as there are other tests with similar. Let me know if you'd be interested in having something like `waitSignal` and I could attempt a crude implementation.

andy-sweet (approved):
Looks good to me!

tests/test_throttler.py
tlambert03: > Always sad to see hopeful sleeps in tests.

yeah, I totally agree.
There's no equivalent (however... `pytest.waitSignal` *does* work just fine with psygnal signals too... since it's got the same API as qt)... I could use that here too, just didn't want to have to bring qt into this :joy:

tlambert03: fwiw, I did test a huge range of sleeps (all the way down to 0) and it continued to work.  so I could probably remove it completely.  mostly it's there to induce a beat, if nothing more

src/psygnal/_throttler.py
tlambert03: correct.  `__doc__` will be `None` on mypyc... (i.e. it won't update the doc on the instance itself)

tests/test_throttler.py
tlambert03: was able to remove the sleep here

PR #234, opened 2023-09-16
fix: emission of events from root validators and extraneous emission of dependent fields

### LINKED ISSUE #233: EventedModel does not emit events for all value changes in a root validator
* psygnal version: `main` at the time of writing (a4a3348fb6f81c3aeefc8a1c723360c9a26db23e)
* Python version: 3.7.16
* Operating System: macOS (Intel x86-64)

### Description

When I define an `EventedModel` with a root validator that coerces some field values based on others, changes in derived/constrained values do not trigger events to be emitted.

Coming here from [an issue on napari](https://github.com/napari/napari/pull/4819#issuecomment-1703093535) where similar problems with its similar `EventedModel` cause problems for `Dims`. No fix for that yet.

### What I Did

Here's the simplest and also somewhat artificial reproducer I could write, where a model has two ints (`x` and `y`) and the root validator ensures `x >= y`.

```python
from pydantic import root_validator

class Model(EventedModel):
    x: int
    y: int

    class Config:
        validate_assignment = True

    @root_validator
    def check(cls, values):
        x = values['x']
        if values['y'] > x:
            values['y'] = x
        return values

m = Model(x=2, y=1)
m.events.x.connect(lambda x: print(f'x is now {x}'))
m.events.y.connect(lambda y: print(f'y is now {y}'))
print(m)  # => Model(x=2, y=1)
m.x = 0   # x change is printed, but not y
print(m)  # => Model(x=0, y=0)
```

I think this is a bug because I'd expect the change in `y` to emit an event.

### Thoughts

I'm not sure this is a critical bug because I've at least had half-baked thoughts about alternatives/workarounds (e.g. define `x` and `y` as properties and then use `property_dependencies`). Taking that half-baked idea a little further, does it make any sense to allow `property_dependencies` to include `fields` as well as properties?

Other ideas/workarounds are welcome!

tlambert03:
fixes #233 

@andy-sweet, this implements option number 2 from https://github.com/pyapp-kit/psygnal/issues/233#issuecomment-1722301583

- `property_dependencies` can include any field names ... not just property setters
- values of dependents are checked before and after setting and events are only emitted if necessary

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -21,7 +21,7 @@ jobs:
       run:
         shell: bash -l {0}
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           fetch-depth: 0
       - uses: conda-incubator/setup-miniconda@v2
--- a/.github/workflows/pypi.yml
+++ b/.github/workflows/pypi.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           fetch-depth: 0
 
@@ -60,7 +60,7 @@ jobs:
             macos_arch: "arm64"
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           fetch-depth: 0
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,7 +16,7 @@ jobs:
     name: Check Manifest
     runs-on: ubuntu-latest
     steps:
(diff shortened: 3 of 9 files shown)


andy-sweet (approved):
Looks good to me. Thanks for the fast fix. Just made a couple of nit-picking comments on implementation comment naming, which are optional.

andy-sweet:
As discussed in #233, it probably also makes sense to rename `property_dependencies` with this solution. That doesn't have to happen in this PR, but there should be an issue to track it if this is merged without.

tlambert03:
agreed, did you have a thought on the exact name?  As mentioned over there:

> I like `dependencies`, but since the namespace is shared with all the other pydantic stuff, perhaps `event_dependencies` is safer/clearer? or do you like `field_dependencies` better than `event_dependencies`?

PR #255, opened 2024-01-25
fix: add and fix copy operators

### LINKED ISSUE #254: Copy operator of evented sets lead to sharing inner structures between copies
### Original issue
https://github.com/napari/napari/issues/6620

### Description

When debugging linked issue, I noticed that `copu` of `pysygnal.containers.Selection` shares internal structures with the original one. As I understand the code of the copy operator, it does only shallow copy of internal structures. 

https://github.com/pyapp-kit/psygnal/blob/b0d2057025e6b8b9facc0210d01222bfdc356a8a/src/psygnal/containers/_evented_set.py#L92-L95

There is any reason to not implement copy as:

```
    def __copy__(self: _Cls) -> _Cls:
        return self.copy()
```

Where `copy()` is implemented as:

https://github.com/pyapp-kit/psygnal/blob/b0d2057025e6b8b9facc0210d01222bfdc356a8a/src/psygnal/containers/_evented_set.py#L97-L98

Current copy may also copy callbacks connected to signals.

Czaki:
closes #254

This PR fixes copy of EventedSet and add copy operator to EventedList and EventedDic.

### DIFF
--- a/src/psygnal/containers/_evented_dict.py
+++ b/src/psygnal/containers/_evented_dict.py
@@ -1,13 +1,12 @@
 """Dict that emits events when altered."""
+from __future__ import annotations
 
 from typing import (
     TYPE_CHECKING,
-    Dict,
     Iterable,
     Iterator,
     Mapping,
     MutableMapping,
-    Optional,
     Sequence,
     Tuple,
     Type,
@@ -42,13 +41,13 @@ class TypedMutableMapping(MutableMapping[_K, _V]):
 
     def __init__(
         self,
-        data: Optional[DictArg] = None,
+        data: DictArg | None = None,
         *,
         basetype: TypeOrSequenceOfTypes = (),
         **kwargs: _V,
     ):
-        self._dict: Dict[_K, _V] = {}
-        self._basetypes: Tuple[Type[_V], ...] = (
+        self._dict: dict[_K, _V] = {}
+        self._basetypes: tuple[type[_V], ...] = (
             tuple(basetype) if isinstance(basetype, Sequence) else (basetype,)
         )
         self.update({} if data is None else data, **kwargs)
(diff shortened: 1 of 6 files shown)


src/psygnal/containers/_evented_dict.py
```diff
@@ -76,19 +77,20 @@ def _type_check(self, value: _V) -> _V:
...
-    ) -> "TypedMutableMapping[_K, _V]":
+    def __newlike__(self: _Cls, mapping: MutableMapping[_K, _V]) -> _Cls:
         new = self.__class__()
         # separating this allows subclasses to omit these from their `__init__`
         new._basetypes = self._basetypes
         new.update(mapping)
         return new
 
-    def copy(self) -> "TypedMutableMapping[_K, _V]":
+    def copy(self: _Cls) -> _Cls:
```
tlambert03, line 87: not necessary if you don't want to but i think this is where `typing.Self` is useful?
https://docs.python.org/3/library/typing.html#typing.Self

Czaki: `typing.Self` is available since python 3.11. I know that it is available from `typing_extensions`. But I remember some problems. Especially when psygnal support python 3.7

tlambert03: it would go in the `if TYPE_CHECKING` section.  psygnal supports 3.7 at runtime, but not for type checking.  We  use *many* recent typing constructs (including `ParamSpec`) that are only supported in later versions of python, so as long as you restrict your import to `TYPE_CHECKING` it's fine

Czaki: Changed

tlambert03 (approved):
thanks again

PR #260, opened 2024-02-01
Name conflict in SignalGroup

getzze:
* psygnal version: 0.9.5
* Python version: 3.11

### Description

I created an Evented Dataclass with private field names in conflict with the attributes of the `SignalInstance` class.
This resulted in an AttributeError. The bug comes from a name conflicts inside `SignalGroup`.

I tried an easy solution to avoid name conflict, by allowing to set a suffix for the signal names.

### What I Did

```python
from typing import ClassVar
from attrs import define, field
from psygnal import SignalGroupDescriptor, Signal, SignalGroup

# Evented Dataclass error
@define
class Test:
    events: ClassVar[SignalGroupDescriptor] = SignalGroupDescriptor()

    _name: str = field(default="name")
#    _args_queue: int = field(default=1)
#    _instance: int = field(default=1)
    age: int = field(default=1)

m = Test()
m.events  # -> AttributeError: 'str' object has no attribute 'connect'

# SignalGroup error
class Events(SignalGroup):
    sig1 = Signal(int)
    _name = Signal(int)

events = Events()  # ->  AttributeError: 'str' object has no attribute 'connect'

# With this PR
@define
class Test:
    events: ClassVar[SignalGroupDescriptor] = SignalGroupDescriptor(signal_suffix="_changed")

    _name: str = field(default="name")
    _args_queue: int = field(default=1)
    _instance: int = field(default=1)
    age: int = field(default=1)

m = Test()
@m.events.connect
    field_name = info.signal.name.removesuffix("_changed")
    print(f"signal {info.signal.name!r} emitted because the field {field_name!r} changed: {info.args}")

m._name = "new" 
# >> signal '_name_changed' emitted because the field '_name' changed: ('new', 'name')'_name_changed' changed to ('new', 'name')

```

### DIFF
--- a/src/psygnal/_evented_decorator.py
+++ b/src/psygnal/_evented_decorator.py
@@ -32,6 +32,7 @@ def evented(
     equality_operators: Optional[Dict[str, EqOperator]] = None,
     warn_on_no_fields: bool = ...,
     cache_on_instance: bool = ...,
+    signal_suffix: str = ...,
 ) -> T:
     ...
 
@@ -44,6 +45,7 @@ def evented(
     equality_operators: Optional[Dict[str, EqOperator]] = None,
     warn_on_no_fields: bool = ...,
     cache_on_instance: bool = ...,
+    signal_suffix: str = ...,
 ) -> Callable[[T], T]:
     ...
 
@@ -55,6 +57,7 @@ def evented(
     equality_operators: Optional[Dict[str, EqOperator]] = None,
     warn_on_no_fields: bool = True,
     cache_on_instance: bool = True,
+    signal_suffix: str = "",
 ) -> Union[Callable[[T], T], T]:
     """A decorator to add events to a dataclass.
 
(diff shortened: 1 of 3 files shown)


getzze:
I tried modifying SignalGroup to store the SignalInstance's in a `signal_instances` dict instead but I didn't manage to make it work, also `self.events.field` was not working anymore.

So I made this minimal change instead.

getzze:
Ok, now we make sure that the `signal` variable in `_setattr_and_emit_` is really a `SignalInstance` and not a str or whatever if we catch an attribute of SignalInstance whose name conflicts with the field name.

I also added another keyword argument to convert private fields to alias or skip them, so it solves #261.

getzze:
I reverted the stuff about private fields as it is not related to the problem of this PR

getzze:
Nice!
I think 3 would be best, it gives more flexibility. But the most important is 2, otherwise everything is totally messed up.
I think the way Signal are attached to SignalGroup should be changed:
- the signals should be stored in a protected attribute, like `__signal_instances`, in SignalGroup instances.
- normal attributes inherited from SignalInstance stay in `__dict__`
- SignalGroup has 2 methods: 
* `get_signal` to return the SignalInstance associated with the name.
* `get_safe_attr` to return the normal attribute from SignalInstance.
- we overwrite `__getattr__` to look if the name was found in `__signal_instances` first, and return it. Otherwise return the normal attribute.
Like this we can get the signals and the normal attributes safely, and `self.events.x` will return the signal also.


I don't know what you think about it.

getzze:
Maybe it was not very clear, but I also meant that in `__init_subclasses` the Signals should be deleted from the `__dict__` of the class.

PR #262, opened 2024-02-02
fix: avoid name conflicts (option 2) use __get__ directly to create signal instances

tlambert03:
alternative to #260, for discussion

### DIFF
--- a/src/psygnal/_group.py
+++ b/src/psygnal/_group.py
@@ -113,7 +113,7 @@ class SignalGroup(SignalInstance):
             name=name or self.__class__.__name__,
         )
         self._sig_was_blocked: dict[str, bool] = {}
-        for _, sig in self.signals.items():
+        for sig in self.signals.values():
             sig.connect(self._slot_relay, check_nargs=False, check_types=False)
 
     def __len__(self) -> int:
@@ -122,7 +122,19 @@ class SignalGroup(SignalInstance):
     @property
     def signals(self) -> dict[str, SignalInstance]:
         """Return {name -> SignalInstance} map of all signal instances in this group."""
-        return {n: getattr(self, n) for n in type(self)._signals_}
+        # Here we convert all of the `Signal`s defined at the class level
+        # into `SignalInstance`.
+        # (these signals were previously collected in `__init_subclass__`)
+        # However, it's a bit more complicated in order to avoid name conflicts.
+        # Rather than using the normal descritor protocol with `getattr(self, name)`,
+        # we *directly* use the `__get__` method of the descriptor, just in case
+        # the signal name conflicts with a method or attribute on this SignalGroup.
+        # this means that signal names will always take priority over other attributes
+        # in a signal group
+        cls = type(self)
+        return {
+            name: signal.__get__(self, cls) for name, signal in cls._signals_.items()
+        }
 
     @classmethod
     def is_uniform(cls) -> bool:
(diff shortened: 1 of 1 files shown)


getzze:
Follow up on https://github.com/pyapp-kit/psygnal/pull/260#issuecomment-1924467493 
This first idea I had would be too complicated to implement I think because we need to mess with `__getattribute__` and it's very error prone. So I was thinking that `SignalGroup` could be a container class instead.

Positive:
* it would really minimize the name conflicts, we could even track the potential conflicts and raise an error so it would be completely conflict-free!
* this redesign would make it easier to call nested signals, e.g. defining `__getitem__`, with `events["foo.sig"]`

Negative
* writing `events.connect` would be longer, `events.agg.connect`
* it would need a big rewrite, although we could create a new class, for instance `SignalContainer`, and do not touch `SignalGroup` so people could choose to use one or the other.


```python
# Rough implementation, I have to figure out the exact layout because of the descriptors

class SignalAggInstance(SignalInstance):
    def _slot_relay(self, *args: Any) -> None:
        pass

    ...


class SignalAgg(Signal):
    _signal_instance_class: SignalAggInstance

# Container for Signal descriptors
class SignalList:
    pass


class SignalContainer:
    _agg_: ClassVar[SignalAgg]
    _signals_: ClassVar[SignalList]
    _uniform_: ClassVar[bool] = False
    _signal_aliases_: ClassVar[Mapping[str, str | None]]

    def __init_subclass__(
        cls,
        strict: bool = False,
        signal_aliases: Mapping[str, str] = {},
    ) -> None:
        """Finds all Signal instances on the class and add them to `cls._signals_`."""
        all_signals = {}
        for k in dir(cls):
            v = getattr(cls, k)
            if isinstance(v, Signal):
                all_signals[k] = v
                # delete from dir
                delattr(cls, k)

        cls._signals_ = type(f"{cls.__name__}List", (SignalList, ), all_signals)

        cls._uniform_ = _is_uniform(cls._signals_.values())
        if strict and not cls._uniform_:
            raise TypeError(
            )

        cls._signal_aliases_ = {**signal_aliases}

        # Create aggregated signal
        cls._agg_ = Signal(...)

        return super().__init_subclass__()

    def __init__(self):
        self.__agg__attribute_name__ == "agg"
        for k, v in self._signal_aliases_.items():
            if v == "__agg__":
                self.__agg__attribute_name__ == k
                break

    def get_aggregated_signals(self):
        return self._agg_

    def get_signal(self, name: str):
        return getattr(self._signals_, name, None)

    def __getattr__(self, name: str, owner):
        sig = self.get_signal(name)
        if isinstance(sig, SignalInstance):
            return sig
        if name == self.__agg__attribute_name__:
            return self.get_aggregated_signals()
        raise AttributeError


class Events(SignalContainer):
    sig1 = Signal(str)
    sig2 = Signal(str)

events = Events()

def some_callback(record):
    record.signal  # the SignalInstance that emitted
    record.args    # the args that were emitted

events.agg.connect(some_callback)
events.sig1.connect(print)
events.sig2.connect(print)

```

tlambert03:
i can *definitely* see the merit of deprecating/breaking the pattern of using the `SignalGroup` itself as a subclass, and instead requiring a more explicit `events.agg.connect` when you specifically want to listen for any event in the group.  I think that could be well worth the trouble, so I'm very open to it.

Given the size of the change, I will want to ping some interested parties.  I think napari is moving towards fully swapping out their events with psygnal, and we'll want to involve some of them as well.  

So, consider me enthused and intrigued, but will need more time to go through these proposals, summarize them, and get opinions.  Thanks again for you all your time!

getzze:
One thing that could be done as a quick fix also is defining the `__set__` method for `Signal` to make it a data descriptor.
Then it gets precedence over normal attributes. Like this it raises an early error so it's easier to understand the problem and debug:
https://docs.python.org/3/howto/descriptor.html#descriptor-protocol


```python
class Signal:
    ...

    def __set__(self, instance: Any, value: Any) -> None:
        """Define __set__ to make it a data descriptor.

        A data descriptor takes precedence over normal attributes with the same name.
        """
        raise AttributeError(
            f"Setting SignalInstance {self._name!r} to a new value is not allowed.\n"
            f"Maybe the signal name conflicts with an attribute of {type(instance)}"
        )
    ...

```

tlambert03:
closing this in favor of #269

PR #266, opened 2024-02-09
ci: inherit secrets in reusable workflow

tlambert03:


tlambert03:
didn't do what I want... but might be because it's a PR.  merging to test

PR #269, opened 2024-02-12
refactor!: New SignalGroup that does not subclass SignalInstance

### LINKED ISSUE #267: Proposal for alternate SignalGroup pattern
This is meta-issue collecting thoughts and proposals about solving name conflicts in SignalGroups. 
*(see also)*
* #260
* #261
* #262
* #265 

The primary problem stems from the fact that 
1. `SignalGroup` inherits from `SignalInstance` (which has a fair number of attributes, like `name`, etc...)
2. Signals names within a `SignalGroup` are determined by the user, and accessible as attributes 
3. This leads to the possibility of naming collisions (such as `SignalGroup().name`)

As a little history, the pattern of inheritance was originally borrowed/inspired from vispy's `EmitterGroup` (akin to `SignalGroup`, which inherits from `EventEmitter` (akin to `SignalInstance`).  I'll note that vispy [simply forbids](https://github.com/vispy/vispy/blob/5ca09c8964f72708aeda342366f6bd84dcb76240/vispy/util/event.py#L655-L658) you to use a name that conflicts.  But I don't think that's an acceptable approach for psygnal, particularly if we want to support evented dataclasses (which are almost certainly going to have fields like `name` at some point).

At the moment, I'm leaning towards a new class entirely, as proposed by @getzze in https://github.com/pyapp-kit/psygnal/issues/262#issuecomment-1936111492 and copied below.  This new class would be a plain container with minimal private attributes, preventing name conflicts.  The primary thing we lose there by not inheriting from `SignalInstance` is that you can't connect to the group class itself using, for example `events.connect`... but that's easily solveable by either adding an aggregating signal (e.g. `events.agg.connect()`), or a special method (e.g. `events.connect_all()` ... risking name conflicts)

------
_Originally posted by @getzze in https://github.com/pyapp-kit/psygnal/issues/262#issuecomment-1936111492_

I was thinking that `SignalGroup` could be a container class instead.

Positive:
* it would really minimize the name conflicts, we could even track the potential conflicts and raise an error so it would be completely conflict-free!
* this redesign would make it easier to call nested signals, e.g. defining `__getitem__`, with `events["foo.sig"]`

Negative
* writing `events.connect` would be longer, `events.agg.connect`
* it would need a big rewrite, although we could create a new class, for instance `SignalContainer`, and do not touch `SignalGroup` so people could choose to use one or the other.

```python
# Rough implementation, I have to figure out the exact layout because of the descriptors

class SignalAggInstance(SignalInstance):
    def _slot_relay(self, *args: Any) -> None:
        pass

    ...


class SignalAgg(Signal):
    _signal_instance_class: SignalAggInstance

# Container for Signal descriptors
class SignalList:
    pass


class SignalContainer:
    _agg_: ClassVar[SignalAgg]
    _signals_: ClassVar[SignalList]
    _uniform_: ClassVar[bool] = False
    _signal_aliases_: ClassVar[Mapping[str, str | None]]

    def __init_subclass__(
        cls,
        strict: bool = False,
        signal_aliases: Mapping[str, str] = {},
    ) -> None:
        """Finds all Signal instances on the class and add them to `cls._signals_`."""
        all_signals = {}
        for k in dir(cls):
            v = getattr(cls, k)
            if isinstance(v, Signal):
                all_signals[k] = v
                # delete from dir
                delattr(cls, k)

        cls._signals_ = type(f"{cls.__name__}List", (SignalList, ), all_signals)

        cls._uniform_ = _is_uniform(cls._signals_.values())
        if strict and not cls._uniform_:
            raise TypeError(
            )

        cls._signal_aliases_ = {**signal_aliases}

        # Create aggregated signal
        cls._agg_ = Signal(...)

        return super().__init_subclass__()

    def __init__(self):
        self.__agg__attribute_name__ == "agg"
        for k, v in self._signal_aliases_.items():
            if v == "__agg__":
                self.__agg__attribute_name__ == k
                break

    def get_aggregated_signals(self):
        return self._agg_

    def get_signal(self, name: str):
        return getattr(self._signals_, name, None)

    def __getattr__(self, name: str, owner):
        sig = self.get_signal(name)
        if isinstance(sig, SignalInstance):
            return sig
        if name == self.__agg__attribute_name__:
            return self.get_aggregated_signals()
        raise AttributeError


class Events(SignalContainer):
    sig1 = Signal(str)
    sig2 = Signal(str)

events = Events()

def some_callback(record):
    record.signal  # the SignalInstance that emitted
    record.args    # the args that were emitted

events.agg.connect(some_callback)
events.sig1.connect(print)
events.sig2.connect(print)

```

tlambert03:
This PR implements the idea proposed by @getzze in #267 

Notable changes in this PR:

- `SignalGroup` is no longer a `SignalInstance` subclass.  All methods that used to be accessible on `SignalGroup` (such as `connect`, `disconnect`, `blocked`, etc...) should now be accessed on the signal relay at  `group.all` .
- Signal instances may now be accessed on a group using both `SignalGroup.__getitem__`, in addition to `__getattr__`.
- SignalGroup is sort of like a Mapping, with methods `__iter__` (to list all signal names), `__len__` (number of signals), `__getitem__` (get a signal). But does *not* currently subclass `abc.Mapping` (so it does not have `items()`, `values()`, `keys()`)

Deprecations:
- accessing any SignalInstance name (e.g. `connect`, etc) on a SignalGroup is deprecated, (but still works for now).  Using `group.all` instead.
- accessing `signals` on a group instance (to get a dict of name->signal instance) is deprecated. Instead use a combination of Mapping methods `__iter__` and `__getitem__`/`__getattr__` 
 
Breaking:
- Because `SignalGroup` no longer inherits `SignalInstance`, the semantics of the `__len__` method on a signal group changes; it now returns the number of signal names, *not* the number of connected slots.  This is probably the main breaking change that I can't deprecate.

closes #260 
closes #267


tlambert03:
@getzze, if you have a moment to take a look, I would appreciate it!  This implements your proposal from https://github.com/pyapp-kit/psygnal/pull/262#issuecomment-1936111492 

It does not yet add aliases or prefixes as you've done in other PRs (you can do those in another PR if you still want them after this)

tlambert03:
one open question about the name `all` for the SignalRelay (the one that can be used to connect everything with `events.all.connect`).  Should we just a clunkier but always guaranteed public name like `events.all_psygnals`?

getzze:
I had a look, but I was waiting for you to "finish" it before trying it.
I think having the possibility to "getitem"s is great!

tlambert03:
thanks!  Yep, I think it's approaching stability and ready for review now.

getzze:
Because we have the dict-like access, I was thinking to dedicate one key to the relay, like `events["__all__"]`. What do you think?

tlambert03:
> I was thinking to dedicate one key to the relay, like `events["__all__"]`. What do you think?

i think i'm (mildly) against it at this point.  reasons being:
- we already have both a private and public way to access that object.  so it's doesn't add anything, just creates additional API surface.
- it raises the question of whether the `'__all__'` key should be included in the keys yielded by `__iter__`, and i think both including it and excluding it are confusing:
    - if we include it in `__iter__`, then we should also probably make `__len__` return +1 (otherwise `len(self) != len(list(self))`... and i think that makes both `len` and `iter` less useful (i can imagine needing to now add a conditional to check whether the key is the special key or not
    - if we don't include it in `__iter__` (which would be my preference) then the only way one would know about it is to "just know about it"... which is not detrimental *per se* ... but it also doesn't seem all that helpful given that it doesn't add new functionality.
- as a string value passed to `__getitem__` , it's harder (though not impossible) to catch in refactoring and stuff

getzze:
You are right that having a "__all__" key would make the behavior of `iter` inconsistent. I agree with you that we should not have it then.

tlambert03:
my inclination is to merge this and release a pre-release candidate that sits for a week or two.
@getzze, if you would like time to have a closer/longer look at this PR before merge, just let me know (no hurry at all) and I'll wait on that.
@Czaki and @jni, if either of you wanted a final chance to comment on the implementation, here ya go

src/psygnal/_group.py
```diff
@@ -40,95 +42,45 @@ class EmissionInfo(NamedTuple):
...
-    def is_uniform(cls) -> bool:
-        """Return true if all signals in the group have the same signature."""
-        return cls._uniform
+        # silence any warnings about failed weakrefs (will occur in compiled version)
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")
+            for sig in signals.values():
+                sig.connect(
+                    self._slot_relay, check_nargs=False, check_types=False, unique=True
+                )
```
Czaki, line 83: Maybe add to `connect` parameter that will allow silence warning, or provide a private context manager for this. 

(If I correctly understand that compiled version will emit warning even with this operator).

src/psygnal/_group.py
```diff
@@ -40,95 +42,45 @@ class EmissionInfo(NamedTuple):
     args: tuple[Any, ...]
 
 
-@mypyc_attr(allow_interpreted_subclasses=True)
-class SignalGroup(SignalInstance):
-    """`SignalGroup` that enables connecting to all `SignalInstances`.
+class SignalRelay(SignalInstance):
```
Czaki, line 45: Create variant that allow to connect to not deprecated signals?

src/psygnal/_group.py
```diff
@@ -179,8 +131,8 @@ def connect_direct(
         """
 
         def _inner(slot: Callable) -> Callable:
-            for sig in self.signals.values():
-                sig.connect(
+            for sig in self._signals:
```
Czaki, line 134: Why is here without catch warnings?

src/psygnal/_group_descriptor.py
```diff
@@ -265,7 +265,12 @@ def _setattr_and_emit_(self: object, name: str, value: Any) -> None:
             group: SignalGroup | None = getattr(self, signal_group_name, None)
             signal: SignalInstance | None = getattr(group, name, None)
```
getzze, line 266: with a dataclass `name` field, `getattr(group, name, None)` returns `''` because of the deprecated looking for `SignalRelay` attributes.
Then `signal` is a `str` , leading to errors

This works:
```python
if group is None or not isinstance(group, SignalGroup):
    return super_setattr(self, name, value)
signal: SignalInstance | None = group[name] if name in group else None
```

src/psygnal/_group_descriptor.py
```diff
@@ -265,7 +265,12 @@ def _setattr_and_emit_(self: object, name: str, value: Any) -> None:
...
             # don't emit if the signal doesn't exist or has no listeners
-            if group is None or signal is None or len(signal) < 2 and not len(group):
+            if (
+                group is None
+                or signal is None
```
getzze, line 270: `isinstance(signal, SignalInstance)`

src/psygnal/_group.py
```diff
@@ -237,17 +190,170 @@ def disconnect(self, slot: Callable | None = None, missing_ok: bool = True) -> N
...
+        return len(self._psygnal_signals)
+
+    def __getitem__(self, item: str) -> SignalInstance:
+        """Get a signal instance by name."""
+        return self._psygnal_instances[item]
+
+    def __iter__(self) -> Iterator[str]:
+        """Yield the names of all signals in the group."""
+        return iter(self._psygnal_signals)
+
```
getzze, line 339:
```python
    def __contains__(self, item: str) -> bool:
        """Check that the signal name exists"""
        return item in self._psygnal_instances
```

src/psygnal/_group.py
```diff
@@ -237,17 +190,170 @@ def disconnect(self, slot: Callable | None = None, missing_ok: bool = True) -> N
...
+        setattr(self, self._psygnal_relay_name, self._psygnal_relay)
+
+    def __init_subclass__(cls, strict: bool = False) -> None:
+        """Collects all Signal instances on the class under `cls._psygnal_signals`."""
+        cls._psygnal_signals = {
+            k: val
+            for k, val in getattr(cls, "__dict__", {}).items()
+            if isinstance(val, Signal)
+        }
+
```
getzze, line 285: Add a mechanism to catch potential name conflicts early on.
The reserved attributes can be the one starting with `_psygnal_` or a reduced list (in this case we may have conflicts if we add private attributes to SignalGroup later.

```python
        conflict_keys = set(k for k in cls._psygnal_signals if k.startswith("_psygnal_"))
        if len(conflict_keys) > 0:
            warnings.warn(
                "Signal names are conflicting with SignalGroup private attribute, "
                f"skip signals: {conflict_keys}",
                stacklevel=2,
            )
            for key in conflict_keys:
                del cls._psygnal_signals[key]
```

src/psygnal/_group.py
```diff
@@ -179,8 +131,8 @@ def connect_direct(
         """
 
         def _inner(slot: Callable) -> Callable:
-            for sig in self.signals.values():
-                sig.connect(
+            for sig in self._signals:
+                self._signals[sig].connect(
```
getzze, line 135: Directly iterate over values
```python
for sig in self._signals.values():
    sig.connect(
```

src/psygnal/_group.py
```diff
@@ -237,17 +190,170 @@ def disconnect(self, slot: Callable | None = None, missing_ok: bool = True) -> N
...
+            name: signal.__get__(self, cls)
+            for name, signal in cls._psygnal_signals.items()
+        }
+        self._psygnal_relay = SignalRelay(self._psygnal_instances, instance)
+
+        # determine the public name of the signal relay.
+        # by default, this is "all", but it can be overridden by the user by creating
+        # a new name for the SignalRelay annotation on a subclass of SignalGroup
+        # e.g. `my_name: SignalRelay`
+        self._psygnal_relay_name = "all"
```
getzze, line 270: How can we define a custom name from `SignalGroupDescriptor` ?
A keyword argument to `__init__` ? Allow parsing instanciated `SignalRelay` attributes, so we can pass it with `type("MyGroup", (SignalGroup, ), {"relay": SignalRelay()})` ?

getzze:
I tried pushing the potential conflicts and it works very well :)

```python
from dataclasses import dataclass, field
from typing import ClassVar
from psygnal import SignalGroupDescriptor, EmissionInfo, Signal

@dataclass
class Model:
    events: ClassVar = SignalGroupDescriptor()

    connect: int = 3
    name: str = "name"
    all: float = 1.0

m = Model()
@m.events.all.connect
    print(f"field {info.signal.name!r} changed to {info.args}")

m.events.all  # <SignalRelay on Model(connect=3, name='name', all=1.0)>
m.events.connect  # <_DataclassFieldSignalInstance 'connect' on <SignalGroup 'ModelSignalGroup' with 3 signals>>
m.events["all"]  # <_DataclassFieldSignalInstance 'all' on <SignalGroup 'ModelSignalGroup' with 3 signals>>

m.events.connect(print)  # prints: "field 'connect' changed to (<built-in function print>,)"
# this can be strange as it is equivalent to m.events.connect.emit(print)
# but hey it was not even allowed before.
```

getzze:
Also, it would be good to define a function to returns the `SignalRelay` from a `SignalGroup`, so we have an easy public API to get the relay, an API that would not change if the private attributes of `SignalGroup` are redefined for instance. It would be better if it was a method of `SignalGroup` but the name could confict with signals.

```python
def get_relay(group: SignalGroup) -> SignalRelay:
    return group._psygnal_relay
```

getzze:
Some idea for later (for instance for the nested events): add a method to SignalRelay to merge two SignalRelays:
```python
def update(self, relay: SignalRelay, take_over: bool = True):
    self._signals.update(relay._signals)
    self._sig_was_blocked.update(relay._sig_was_blocked)
    for sig in relay._signals.values():
        if take_over:
            sig.disconnect(relay._slot_relay)
        sig.connect(self._slot_relay, check_nargs=False, check_types=False, unique=True)
```

src/psygnal/_group.py
tlambert03: I'm not following what you mean by allow to connect to not deprecated signals

tlambert03: do you mean a variant that doesn't deprecate direct access to all of the `SignalInstance` names?
If so, I would ask why?  Do you think it's actually a better design?  Or don't want to have to deal with the deprecation? (fwiw, I tested against napari and magicgui, napari is fine, and magicgui is fine after https://github.com/pyapp-kit/magicgui/pull/629).  
I'd rather just decide whether it's deprecated or not, rather than have a variant that is deprecated and one that isn't... but let me know if I misunderstood you there

src/psygnal/_group.py
tlambert03: there are no warnings to catch here.  if you let me know what warnings you were expecting might happen, I'm happy to help clarify

src/psygnal/_group.py
tlambert03: yep, can do this.  thanks. In a previous PR, the `SignalRelay` directly took the `SignalGroup` instance as an input (which didn't have the `values()` attribute).  But then I changed it and didn't update this pattern here.

src/psygnal/_group.py
tlambert03: as described in the comment above and in the class docstring:  as a SignalRelay annotation in the class:

```python
class MyGroup(SignalGroup):
    custom_relay_name: SignalRelay 

# or, if you want to do it dynamically, which I'm less concerned about syntax
type("MyGroup", (SignalGroup, ), {"__annotations__": {'custom_relay_name': SignalRelay} })
```

The reason for doing it this way is that it will satisfy type checkers as well.  If we take an argument in `__init__`, then the end-user won't get any type-hinting or autocompletion for that name. If they use the syntax above, they  will get type checking and completion.

src/psygnal/_group.py
tlambert03: i like it, thanks

src/psygnal/_group_descriptor.py
tlambert03: hah, that's funny, I had [a previous iteration](https://github.com/pyapp-kit/psygnal/pull/269/commits/9f44dfcfbd40b640bec98bbb7fcfb8a0749e814c#diff-8218e71615e3cb1b30f78bd2143ca77661925d39178dd9ccad62dc0f62206c98R227-R228) that would have solved this (by first looking at the signal names in `SignalGroup.__getattr__` before checking if the name is on `SignalInstance`), but then removed it because I forgot why I had put it there.  THIS is why :joy:  I'll add it back and add a test

src/psygnal/_group_descriptor.py
tlambert03: used group.__getitem__ instead, and fixed the __getattr__ issue above that should also fix this

src/psygnal/_group_descriptor.py
tlambert03: > with a dataclass name field, getattr(group, name, None) returns '' because of the deprecated looking for SignalRelay attributes.

actually... @getzze, are you sure about that? If I run the following code on the commit you commented on 8eab8af, i don't get the issue you describe:

```python
from psygnal import evented, SignalGroup, SignalInstance
from dataclasses import dataclass


@evented
@dataclass
class Foo:
    name: str


obj = Foo("foo")
assert obj.name == "foo"
group = obj.events
assert isinstance(group, SignalGroup)
assert "name" in group
assert isinstance(group.name, SignalInstance)
assert group["name"] is group.name
```

I think the explanation is that `getattr(group, name, None)` first calls getattr on the SignalGroup *class* (not the instance), and that finds the Signal descriptor's `__get__` method, and it never reaches `__getattr__` method on the group instance (and you never reach the deprecation warning).  I did have a comment on this in the code afterall [here](https://github.com/pyapp-kit/psygnal/pull/269/files#diff-8218e71615e3cb1b30f78bd2143ca77661925d39178dd9ccad62dc0f62206c98R305-R311).  Did you try this locally and find otherwise?  Or were you just looking at the code online and expecting it to hit the deprecation?

tlambert03: I've added back the following code to `__getattr__` just in case, but it will likely never hit:

```python
        # Note, technically these lines aren't actually needed because of the descriptor
        # protocol.  Accessing a name on the instance will first look in the
        # instance's __dict__, and then in the class's __dict__, which
        # will call Signal.__get__ and return the SignalInstance.
        # these lines are here as a reminder to developers.
        if name != "_psygnal_instances" and name in self._psygnal_instances:
            return self._psygnal_instances[name]  # pragma: no cover
```

tlambert03:
> Also, it would be good to define a function that returns the `SignalRelay` from a `SignalGroup`, so we have an easy public API to get the relay, an API that would not change if the private attributes of `SignalGroup` are redefined for instance. It would be better if it was a method of `SignalGroup` but the name could confict with signals.

yeah, this is what I was getting at in  https://github.com/pyapp-kit/psygnal/pull/269#issuecomment-1941901423

I agree, there should be a guaranteed public property or method.  my inclination would be an `@property`, as the final syntax is nicer, but do you have a preference for a getter method instead?.  And the exact name is the main question.  Something like `.psygnal_relay` is pretty guaranteed to never have a name conflict.
And if we do that, is there perhaps no reason to provide all the logic for letting them override the public name?  Maybe one safe public name is all we need

src/psygnal/_group.py
tlambert03: (linking to https://github.com/pyapp-kit/psygnal/pull/269#issuecomment-1947225024 ... it's possible we don't want this mechanism at all if we're going to provide a permanent public name)

getzze:
```python
type("MyGroup", (SignalGroup, ), {"__annotations__": {'custom_relay_name': SignalRelay}, "name": Signal(str, str) })
```
Then `MyGroup.name` does not have type annotations, we would also need to add the annotation to `"__annotations__"`.

Anyway, if needed, all this boilerplate code can be put in a class method.

src/psygnal/_group_descriptor.py
getzze: > actually... @getzze, are you sure about that? If I run the following code on the commit you commented on [8eab8af](https://github.com/pyapp-kit/psygnal/commit/8eab8afb7b309bac5aec1853276e2718d9bfc620), i don't get the issue you describe:

Maybe the dataclass field was named `_name` not `name`, I cannot test it now.

src/psygnal/_group.py
tlambert03: > Then MyGroup.name does not have type annotations

it *does* have annotations as far as psygnal is concerned... which is simple to look at `cls.__annotations__` in [this one line here](https://github.com/pyapp-kit/psygnal/pull/269/files#diff-8218e71615e3cb1b30f78bd2143ca77661925d39178dd9ccad62dc0f62206c98R271).  But if you mean that mypy and an IDE wouldn't give you annotations, yes that's true, but any dynamic type creation is already hopeless for those things.  
so the point is:  it's always best practice for a user to hint this with a type annotation, and even if we did offer a parameter in `__init__`, they might *still* add that type annotation just for the sake of mypy.  So, we might as well use that mechanism anyway (if we're going to do this at all that is)

tlambert03:
```python
In [1]: class MyGroup(SignalGroup):
   ...:     custom_name: SignalRelay
   ...:

In [2]: Group2 = type("Group2", (SignalGroup, ), {"__annotations__": {'custom_relay_name': SignalRelay}, "name": Signal(str, str) })

In [3]: MyGroup.__annotations__
Out[3]: {'custom_name': <class 'psygnal._group.SignalRelay'>}

In [4]: Group2.__annotations__
Out[4]: {'custom_relay_name': <class 'psygnal._group.SignalRelay'>}

In [5]: MyGroup().custom_name
Out[5]: <SignalRelay>

In [6]: Group2().custom_relay_name
Out[6]: <SignalRelay>
```

tlambert03: > Then MyGroup.name does not have type annotations

oops sorry. I think I misread this the first time.  You're right that 'name' would also need to go in the `__annotations__` dict... but how many dynamically created classes are that worried about type annotations?

getzze:
I was proposing functions and not methods to avoid any name conflict. But if the method name starts with `psygnal` and is long, it's very unlikely that it will conflict with a dataclass field. And a method is more convenient because of auto-completion.

The good thing about `all` is that it's short. Maybe it's not worth putting all the code for allowing changing the `all` name though, and just have a long property/method name. If it's documented what the possible conflicts are and there are very few and unlikely names it should be ok.

We can always add possibility to use aliases, which are useful even after this PR, to alias a private field to a public name.

tlambert03:
thanks @getzze!  yeah I agree.  feels tough to make this rather tiny-but-important decision.  I might ping @jni for his thoughts on this:

Juan, we all like the idea of `events.all.connect` ... but `all` still has the possibility of name conflict (as you pointed out before).  The current PR allows the user to change the name `all` to something else if they want to use it for themselves (thereby avoiding the conflict) ... but then we have the problem of a third party wanting to know how to get at the aggregate/relay connector.  That leads to a desire to have a permanent public name like `events.all_psygnals.connect` or `events.get_relay().connect`, which are uglier, to be sure.  What do you think?  some options:

1. stick with `events.all` by default, but allow the user to pick their own public name, and to hell with anyone else who receives the object :joy:  _(this is probably a no... since magicgui for example needs to know how to get at the relay)_
2. keep `events.all`, and allow user-overrides, *and* add a permanent safe public name like `events.all_psygnals` (name suggestions welcome)
3. use `events.all`, and if someone has a dataclass with a field named `all`, then they *must* access it using `group['all']`. because `events.all` will always point to the relay.
4. don't use `events.all`, just go with one permanent ugly public name `events.all_psygnals` that's unlikely to ever conflict with a dataclass field.

src/psygnal/_group.py
Czaki: I see a similar pattern like in my first comment, but here without catching warnings.

tlambert03: the warning we are specifically connecting in the `__init__` is a failed internal connection of `self._slot_relay`.  That's the only one we want to silence.  Any user-provided callables that are converted into strong references do need to be warned

getzze:
I think I like 3 better.

Somebody mentioned the example of `pandas` that I think is a good model: you can access ALL the columns by key, like `events["name"]`, and by attribute, only the columns with a well-behaved (no space) and non-conflicting name.

src/psygnal/_group.py
tlambert03: just to double check, is it clear to you here that the warning blocked in the context above, and the warnings (i thought) we are discussing here are *not* deprecation warnings?  the warning blocked above is [this one in `_weak_callback`](https://github.com/pyapp-kit/psygnal/blob/94d0a672ad963d3fa5c36d38ce3cc79029f04828/src/psygnal/_weak_callback.py#L232-L235)... 

just wanted to make sure you're not confusing the deprecation warnings that have been added elsewhere in this PR with the ones we're blocking and discussing here

tlambert03:
> you can access ALL the columns by key, like `events["name"]`, and by attribute, only the columns with a well-behaved (no space) and non-conflicting name.

yep, and combined with the warning at class creation that you stubbed out in https://github.com/pyapp-kit/psygnal/pull/269#pullrequestreview-1882868728 ... i think there will be no surprises

jni:
I like 3 also, by a fair margin, so I am happy that we are all in agreement. 😊

tlambert03:
implemented in 47bf3ae

tests/test_evented_decorator.py
```diff
@@ -245,3 +248,28 @@ class Foo:
...
+
+def test_name_conflicts() -> None:
+    # https://github.com/pyapp-kit/psygnal/pull/269
+
+    @evented
+    @dataclass
+    class Foo:
+        name: str
+        all: bool = False
+
```
getzze, line 261: Add more conflicting fields

```python
        is_uniform: bool = True
        signals: list = field(default_factory=list)
       _psygnal_signals: str = "signals"
```

getzze: Mmm, I was expecting `type(group).is_uniform` to be the classmethod but it a `Signal`.
Now that we have a safe way of getting the Signals with `__getitem__` maybe we can keep the few conflicting properties/methods from being overwritten by the Signals. For now there is only `all` and `is_uniform`.
What do you think?

src/psygnal/_group.py
getzze: I think you forgot to add this :)

tlambert03: added thanks

tests/test_evented_decorator.py
tlambert03: i'm going to rename `is_uniform` to `psygnals_uniform`, and deprecate `is_uniform`.  If a dataclass uses it as a name, it takes precedence in all regards.

tlambert03:
ok!  this one is going in :)

Thanks all for the helpful discussion.  Huge thanks to @getzze for getting this name-conflict discussion started in #260, and for proposing the solution that was ultimately implemented here in https://github.com/pyapp-kit/psygnal/pull/262#issuecomment-1936111492.  It's not often that someone comes along and dives deep enough to understand the existing codebase and provides very insightful and productive suggestions.  I hope you don't mind that I ran with that idea and implemented it here (it was a big change that I wanted to consider carefully), and I hope that you'll continue contributing improvements :)

tlambert03:
this will be released as v0.10.0rc0 for a week or so before release

getzze:
Ahahah, likewise, it's not every time that developers include contributors, who know less about the project, to make big changes.

I have others PRs to propose ;)

PR #274, opened 2024-02-20
Add (back) SignalGroup methods

getzze:
While porting the signal aliases to the new `SignalGroup`, I found that it was quite annoying to work with a `SignalGroup`, mainly because it has no methods now.
So I did an experiment, that is to put back methods to `SignalGroup`. Now that the `Signal`s are well separated from the `SignalRelay` it is very easy. I took one step further, removing the `Signal`s from the list of attributes of the `SignalGroup`.
Therefore it is very easy to make two things:
- `SignalInstance`s can always be retrieved using the getter: `group[signal_name]`
- If the signal name does not conflict with any attribute of `SignalGroup`, it can be retrieved as an attribute, like for pandas `DataFrame`s: `group.signal_name`

I created a PR instead of an issue so it is easier to discuss concrete code.

There are other changes in the PR that are just correcting bugs:
- `SignalGroup.__init_subclass__`: use `dir(cls)` instead of `cls.__dict__` to find `Signal`s so it works with sub-classes of (subclass of) `SignalGroup`.
- `__len__`, `__iter__` and `__getitem__` should use `self._psygnal_instances` and not a mix of it and `self._psygnal_signals` for consistency (and if a user modifies one of the dict but not the other). Like this we can have the same interface for the class instead of the instance. Implemented with a `metaclass`.
- add a method to `Signal` to create a `SignalInstance` that can be called to create `SignalGroup._psygnal_instances` instead of using `__get__`. It is also useful to use a `Signal` with a slotted class (like a class created with `attrs.define`).
- modify `utils.monitor_events` to find `Signal`s in a `SignalGroup`.


tlambert03:
> I found that it was quite annoying to work with a SignalGroup, mainly because it has no methods now.

before going down this route to add all these things back, can we dig into this statement a bit?  I guess my question would be what's so annoying?  why is it any more annoying to use `group.all.connect` over `group.connect`?

src/psygnal/_group.py
```diff
@@ -183,12 +192,23 @@ def disconnect(self, slot: Callable | None = None, missing_ok: bool = True) -> N
         super().disconnect(slot, missing_ok)
 
 
+class SignalGroupMetaclass(type):
```
tlambert03, line 195: can you accomplish this without metaclasses?  mypyc doesn't support metaclasses, and in general I've been happy with using mypyc, and wouldn't want to drop it just for this

tlambert03:
> no public API to get `events.signals`, you have to use `{k: events[k] for k in events}`...  instead of `events.get(name, None)` you need to do `events[name] if name in events else None`

i don't find ``{k: events[k] for k in events}` terribly offputting :) but i get the point.  
it's all *close* to a `Mapping`, but it lacks an analog for `items()` and/or `values()`.  I started with an actual Mapping subclass, but removed it due to name conflicts with `values`, `keys` etc...
So, we could either accept these name conflicts, and make it a proper `Mapping` with the interface everyone is used to (but perhaps with names that are more likely to conflict, like `values` or `items` ... Or we could keep it as "mapping-like", and use our own names like `signals` ... or whatever.  But using `signals` to return the dict feels slightly weird, since there's no analog in `Mapping` that simply returns the full map object.

I'm not sure which I prefer there.  you?


> With the new design, the behavior is perfectly controlled, so maybe it's not worth changing too much the original API.
Now we can get the relay and the signals unambiguously, with `events.all` and `events[name]`, the rest of the API could be made simple, intuitive and compatible with the previous version.

I'm definitely open to this, and we should definitely make the decision before I release the release candidate.
I think my first thought is that if we start adding back names, a question arises as to _which_ names, and I kinda dislike "half-way-there" decisions.   For example, in this PR, you've added connect and disconnect, but not `blocked` or many of the other things that people may expect from something that behaves like a SignalInstance.  So, it's harder to build up an intuition for which names one can expect... you just have to keep checking the docs, unless we go "all in" or "all out".  
Between those two (all in vs all out) do you have an opinion?    (Note, I see this question as independent of the previous one, regarding `.get()`, `.values()` etc...)

getzze:
I think the full `Mapping` interface is too much: it would create potential name conflicts and all the part of the interface that are setters would need to be deactivated anyway (you cannot add signals to the instance). So I would go for a minimal "getter" Mapping interface with names that are  meaningful and easy to remember. I agree that `signals` is not a very good name.

Regarding the names that we could allow for `SignalGroup`, I was thinking that users will most probably use adjectives for their signals and due to the data class interface, also names. So attributes like `name` or `is_uniform` will definitely conflict. But it's completely arbitrary I agree.
After thinking about it a bit more, I think all the methods from `SignalRelay` should pass to `SignalGroup`: `dis/connect`, `un/block`, `blocked`, `connect_direct`. Maybe we could also add `emit`, `paused`, `pause`, `resume` from `SignalInstance`.

What do you think?

tlambert03:
>  all the part of the interface that are setters would need to be deactivated anyway (you cannot add signals to the instance). So I would go for a minimal "getter" Mapping interface with names that are  meaningful and easy to remember

`collections.abc.Mapping` _is_ the minimal getter mapping interface, and it adds `__contains__`, `keys`, `items`, `values`, `get`, `__eq__`, and `__ne__`. `collections.abc.MutableMapping` is the one that adds all the setters, and we wouldn't inherit from that one.   see [`collections` docs](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes)

So, as mentioned in https://github.com/pyapp-kit/psygnal/pull/274#issuecomment-1959437371, the main question is whether we want to use `keys`, `items`, `values`, and `get`, and stick with standard mapping API names, or make our own names for these standard getters.

> I think all the methods from `SignalRelay` should pass to `SignalGroup`

I'm ok with that for now.  Easier to deprecate it later than to deprecate it now and walk it back later.  Would you be able to update this PR to that effect?  and remove the metaclass?

getzze:
Ah, I didn't think that `collections.abc.Mapping` was the definition of the interface, sorry :). I think `get` is fine. For `keys`, `items` and `values`, they are conflict-prone but I cannot think of better names, maybe with a "signal_" prefix? They are not great names, so maybe stick with the standard?

I can update the PR yes, but I think you have permissions to modify it also (although I have no idea how you can do that with Github)

tlambert03:
Ok, if you're alright with me thinkering with your branch a bit, I may play around (and will ping you if you need to pull changes).  Thanks!

getzze:
ah sorry, I saw your message after I pushed the changes...

But yes, please modify the PR

getzze:
I made another change, that is to allow signal names starting with "_psygnal_", as you can access them with the `Mapping` interface only. But I don't think we need to remove them altogether.

tlambert03:
> ah sorry, I saw your message after I pushed the changes...

no worry, feel free to keep modifying and pushing.  I can handle any conflicts

src/psygnal/_group.py
```diff
@@ -250,54 +254,58 @@ class MySignals(SignalGroup):
...
-            for k, val in getattr(cls, "__dict__", {}).items()
-            if isinstance(val, Signal)
+        # Collect Signals and remove from class attributes
+        # Use dir(cls) instead of cls.__dict__ to get attributes from super()
+        _psygnal_signals = {}
+        for k in dir(cls):
+            val = getattr(cls, k, None)
+            if isinstance(val, Signal):
+                _psygnal_signals[k] = val
+                delattr(cls, k)
```
tlambert03, line 287: This comment is half note-to-self (explaining something i've just understood) and half note to @getzze  :)

I dislike this particular `delattr(cls, k)` here.  It leads to the rather surprising behavior:

```python
class MyGroup(SignalGroup):
    my_signal = Signal()

hasattr(MyGroup, 'my_signal')  # False
```
Any external observer would be absolutely justified to be rather surprised by that. And I'd like to avoid surprising behavior.

However, I just spent the last couple hour or two to make it work without that, and then I came into the same challenge you presumably encountered with the fact that we (by default) cache `SignalInstances` on the object itself, similar to how `functools.cached_property` works.  That is

```python
my_group()
# at this first access of `my_signal` on `my_group`, the newly created SignalInstance
# is set as an attribute on the my_group instance, overriding the `Signal` on the class
my_group.my_signal 
```

I see now why this PR has to also modify the caching behavior of `Signal.__get__`.  It's because of this one `delattr(cls, k)` line here:  by removing the `Signal` descriptor object, we're free to do our own behavior in `getattr` (which is desire able for controlling name conflicts), but then of course we want to control the caching of that object as well.

I also completely recognize that all of this ugliness stems from the original design of the `SignalGroup` API design itself.  Specifically: while the `Signal` descriptor protocol may works nicely as a class attribute on a regular class, it doesn't work quite as well on `SignalGroup` (inasmuch as we also want that group to have lots of other possible names on it, like `connect`, etc..)

*side note: part of the motivation for the descriptor protocol is that it makes type checking and IDE support much easier.  One thing I dislike about most observer-pattern implementations, is that you never know if you're calling the callback properly... and the descriptor pattern helps us to hint at that during static type checking (even if we haven't yet taken full advantage of that)*

Anyway, after banging my head against this for a little while, I'm again torn as to whether this is a good idea.  We either ask everyone to just use `group.all` (and end up with a clean pattern), or we provide the magical convenience of passing through all of these SignalRelay names, but possibly at the cost of rather surprising behavior for anyone unlucky enough to come into a conflict).  

still thinking...

getzze:
Just adding some thinking:

Actually this works:

```python
g = MyGroup()
hasattr(g, 'my_signal') # True
```

And if `SignalGroup` inherits from a metaclass, this would work:

```python
hasattr(MyGroup, 'my_signal') # True
```
But I think it's not worth losing the `mypyc` speedup. It's just for saying that it could be done.

What could be done is doing `delattr(cls, k)`  just after, when checking for conflict and deleting only the conflicting attributes (keeping all the `Signal`s in `_psygnal_signals` anyway). But we have to be sure that the two types of signals, real attributes and false attributes, are getting initialized identically.


Type checkers don't recognize that `g.my_signal` is a `SignalInstance` if it's not a real attribute?

tlambert03:
> What could be done is doing `delattr(cls, k)`  just after, when checking for conflict and deleting only the conflicting attributes (keeping all the `Signal`s in `_psygnal_signals` anyway). But we have to be sure that the two types of signals, real attributes and false attributes, are getting initialized identically.

yeah, that's the road I was starting to go down.  I agree, it's got potential :)

> Type checkers don't recognize that `g.my_signal` is a `SignalInstance` if it's not a real attribute?

no no... sorry, type checkers are just fine with us using `delattr`, they don't care.  And it's even better actually, because anyone using a protected name, like:

```python
class MyGroup(SignalGroup):
    connect = Signal(int)
```

will get a type error when type checking.  *and* it would work at runtime... Really the only surprising thing is looking for the attribute on the class itself (not on the instance of the class)... which, granted, is going to be rather uncommon.
And, as long as it's only deleted when there's a conflict (as you mentioned above), I think we can probably limit the surprising impact.  But that's the line I'm currently aiming for: only delete class attr in case of conflict, and make sure that actual descriptors and false attributes end up behaving the same way.

tlambert03:
implemented in #286

PR #275, opened 2024-02-20
perf: Fixing performance of evented set

Czaki:
In this PR I have added an option for the combiner function to consume the whole list of tuples at once.

### DIFF
--- a/src/psygnal/_signal.py
+++ b/src/psygnal/_signal.py
@@ -1076,7 +1076,7 @@ class SignalInstance:
 
         Parameters
         ----------
-        reducer : Callable[[tuple, tuple], Any], Callable[[list[tuple]], tuple],  optional
+        reducer : Callable[[tuple, tuple], Any], Callable[[list[tuple]], tuple] | None
             If provided, all gathered args will be reduced into a single argument by
             passing `reducer` to `functools.reduce`.
             NOTE: args passed to `emit` are collected as tuples, so the two arguments
@@ -1134,7 +1134,7 @@ class SignalInstance:
 
         Parameters
         ----------
-        reducer : Callable[[tuple, tuple], Any] or Callable[[list[tuple]], Any], optional
+        reducer : Callable[[tuple, tuple], Any] or Callable[[list[tuple]], Any] | None
             If provided two arguments, all gathered args will
             be reduced into a single argument by
             passing `reducer` to `functools.reduce`.
--- a/src/psygnal/_signal.py
+++ b/src/psygnal/_signal.py
@@ -932,7 +932,9 @@ class SignalInstance:
             If `check_nargs` and/or `check_types` are `True`, and the corresponding
             checks fail.
         """
-        if self._is_blocked or len(self._slots) == 0:
+        if self._is_blocked or (
+            len(self._slots) == 0 and SignalInstance._debug_hook is None
+        ):
             return None
 
         if check_nargs:
--- a/src/psygnal/_signal.py
+++ b/src/psygnal/_signal.py
@@ -18,6 +18,7 @@ from typing import (
     List,
     Literal,
     NoReturn,
+    Tuple,
     Type,
     TypeVar,
     Union,
@@ -45,7 +46,7 @@ if TYPE_CHECKING:
     from ._weak_callback import RefErrorChoice
 
     # single function that does all the work itself
-    ReducerOneArg = Callable[[Iterable[tuple[Iterable, Iterable]]], tuple[tuple, tuple]]
+    ReducerOneArg = Callable[[Iterable[Tuple[Iterable, Iterable]]], Tuple[tuple, tuple]]
     # function that works on two arguments and will be passed to itertools.reduce
     ReducerTwoArgs = Callable[[tuple, tuple], tuple]
     ReducerFunc = Union[ReducerOneArg, ReducerTwoArgs]
--- a/src/psygnal/_signal.py
+++ b/src/psygnal/_signal.py
@@ -24,7 +24,7 @@ from typing import (
     get_args,
     get_origin,
     get_type_hints,
-    overload,
+    overload, List,
 )
 
 from ._exceptions import EmitLoopError
@@ -1111,7 +1111,7 @@ class SignalInstance:
                 hasattr(reducer, "__annotations__")
                 and len(reducer.__annotations__) == 2
             ):
-                args = cast(Callable[[list[tuple]], tuple], reducer)(self._args_queue)
+                args = cast(Callable[[List[tuple]], tuple], reducer)(self._args_queue)
             elif initial is _NULL:
                 args = reduce(
                     cast(Callable[[tuple, tuple], tuple], reducer), self._args_queue
--- a/src/psygnal/_signal.py
+++ b/src/psygnal/_signal.py
@@ -33,7 +33,6 @@ from ._queue import QueuedCallback
 from ._weak_callback import (
     StrongFunction,
     WeakCallback,
-    WeakMethod,
     WeakSetattr,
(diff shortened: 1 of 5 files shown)


Czaki:
When developing locally, `pre-commit` complains about line length, but I do not know how to break type declaration.

Also, mypy report problems with lack of Self (I see that on Ci, it does not happen). 

I need to add type casting for reducer to satisfy mypy.

Czaki:
(Should I add benchmarks to `tests/test_bench`)?

src/psygnal/_signal.py
```diff
@@ -1105,10 +1106,21 @@ def resume(self, reducer: ReducerFunc | None = None, initial: Any = _NULL) -> No
         if not getattr(self, "_args_queue", None):
             return
         if reducer is not None:
-            if initial is _NULL:
-                args = reduce(reducer, self._args_queue)
+            if (
+                hasattr(reducer, "__annotations__")
+                and len(reducer.__annotations__) == 2
+            ):
+                args = cast(Callable[[List[tuple]], tuple], reducer)(self._args_queue)
```
tlambert03, line 1113: this is my only concern at the moment.  I can break tests, for example, just by removing the type annotation from the reduce_events

```python
def _reduce_events(li: list[tuple]): ...  # works
def _reduce_events(li): ...  # doesn't work
```

that's a bit too flimsy of a contract

tlambert03:
> (Should I add benchmarks to tests/test_bench)?

it's ok.  the asv benchmarks are nice because they can run retrospectively (and show the speedup you gained here), whereas codspeed and `test_bench` can only start measuring going forward.  I like codspeed in general, but the lack of retrospective monitoring is a bummer, so the one you added to asv is good enough

src/psygnal/_signal.py
```diff
@@ -1105,10 +1106,21 @@ def resume(self, reducer: ReducerFunc | None = None, initial: Any = _NULL) -> No
...
+                args = reduce(
+                    cast(Callable[[tuple, tuple], tuple], reducer), self._args_queue
+                )
             else:
-                args = reduce(reducer, self._args_queue, initial)
+                args = reduce(
+                    cast(Callable[[tuple, tuple], tuple], reducer),
+                    self._args_queue,
+                    initial,
+                )
```
tlambert03, line 1123:
```suggestion
            if len(inspect.signature(reducer).parameters) == 1:
                reducer = cast(Callable[[Iterable[tuple]], tuple], reducer)
                args = reducer(self._args_queue)
            else:
                reducer = cast(Callable[[tuple, tuple], tuple], reducer)
                if initial is _NULL:
                    args = reduce(reducer, self._args_queue)
                else:
                    args = reduce(reducer, self._args_queue, initial)
```

are you ok with this instead?

src/psygnal/_signal.py
```diff
@@ -43,7 +44,7 @@
     from ._group import EmissionInfo
     from ._weak_callback import RefErrorChoice
 
-    ReducerFunc = Callable[[tuple, tuple], tuple]
+    ReducerFunc = Union[Callable[[tuple, tuple], tuple], Callable[[List[tuple]], tuple]]
```
tlambert03, line 47:
```suggestion
    # single function that does all the work itself
    ReducerOneArg = Callable[[Iterable[tuple[Iterable, Iterable]]], tuple[tuple, tuple]]
    # function that works on two arguments and will be passed to itertools.reduce
    ReducerTwoArgs = Callable[[tuple, tuple], tuple]
    ReducerFunc = Union[ReducerOneArg, ReducerTwoArgs]
```

src/psygnal/containers/_evented_set.py
```diff
@@ -232,22 +232,30 @@ def __init__(self, iterable: Iterable[_T] = ()):
 
     def update(self, *others: Iterable[_T]) -> None:
         """Update this set with the union of this set and others."""
-        with self.events.items_changed.paused(_reduce_events, ((), ())):
+        with self.events.items_changed.paused(
+            _reduce_events,
+        ):
```
tlambert03, line 237:
```suggestion
        with self.events.items_changed.paused(_reduce_events):
```

src/psygnal/containers/_evented_set.py
```diff
@@ -232,22 +232,30 @@ def __init__(self, iterable: Iterable[_T] = ()):
...
             super().update(*others)
 
     def clear(self) -> None:
         """Remove all elements from this set."""
-        with self.events.items_changed.paused(_reduce_events, ((), ())):
+        with self.events.items_changed.paused(
+            _reduce_events,
+        ):
```
tlambert03, line 244:
```suggestion
        with self.events.items_changed.paused(_reduce_events):
```

src/psygnal/containers/_evented_set.py
```diff
@@ -232,22 +232,30 @@ def __init__(self, iterable: Iterable[_T] = ()):
...
             super().clear()
 
     def difference_update(self, *s: Iterable[_T]) -> None:
         """Remove all elements of another set from this set."""
-        with self.events.items_changed.paused(_reduce_events, ((), ())):
+        with self.events.items_changed.paused(
+            _reduce_events,
+        ):
```
tlambert03, line 251:
```suggestion
        with self.events.items_changed.paused(_reduce_events):
```

src/psygnal/containers/_evented_set.py
```diff
@@ -232,22 +232,30 @@ def __init__(self, iterable: Iterable[_T] = ()):
...
             super().difference_update(*s)
 
     def intersection_update(self, *s: Iterable[_T]) -> None:
         """Update this set with the intersection of itself and another."""
-        with self.events.items_changed.paused(_reduce_events, ((), ())):
+        with self.events.items_changed.paused(
+            _reduce_events,
+        ):
```
tlambert03, line 258:
```suggestion
        with self.events.items_changed.paused(_reduce_events):
```

src/psygnal/containers/_evented_set.py
```diff
@@ -299,8 +307,10 @@ def __init__(self, iterable: Iterable[_T] = ()):
...
 
 
-def _reduce_events(a: tuple, b: tuple) -> tuple[tuple, tuple]:
-    """Combine two events (a and b) each of which contain (added, removed)."""
-    a0, a1 = a
-    b0, b1 = b
-    return (a0 + b0, a1 + b1)
+def _reduce_events(li: list[tuple]) -> tuple[tuple, tuple]:
+    """Combine multiple events into a single event."""
+    added_li, removed_li = [], []
```
tlambert03, line 312:
```suggestion
def _reduce_events(li: Iterable[tuple[Iterable, Iterable]]) -> tuple[tuple, tuple]:
    """Combine multiple events into a single event."""
    added_li: list = []
    removed_li: list = []
```

src/psygnal/_signal.py
Czaki: Yes. I wrote more complex cod because of performance worries

src/psygnal/containers/_evented_set.py
Czaki: Lovely black ..

tlambert03: 😂

src/psygnal/_signal.py
Czaki: I needto apply it manually, so I will do it later

tlambert03: ok, thanks!

src/psygnal/_signal.py
tlambert03: (looks like my `tuple[]` broke mypyc compilation on 3.8 ... sorry)

PR #277, opened 2024-02-21
perf: Delay SignalRelay connection to when a callback is connected

### LINKED ISSUE #276: [idea] Connect SignalRelay to signals only when have callback
### Description

Currently, SignalRelay is always connected to all signals in SignalGroup. It means that emitting an event always triggers some callback. 

Maybe SignalRelay should connect itself to all signals in a group only if something is connected to it itself?

tlambert03:
closes #276

### DIFF
--- a/src/psygnal/_group.py
+++ b/src/psygnal/_group.py
@@ -19,6 +19,7 @@ from typing import (
     ContextManager,
     Iterable,
     Iterator,
+    Literal,
     Mapping,
     NamedTuple,
 )
@@ -69,8 +70,8 @@ class SignalRelay(SignalInstance):
 
     def _append_slot(self, slot: WeakCallback) -> None:
         super()._append_slot(slot)
-        if len(self._slots) == 1
-	        self._connect_relay():
+        if len(self._slots) == 1:
+            self._connect_relay()
 
     def _connect_relay(self) -> None:
         # silence any warnings about failed weakrefs (will occur in compiled version)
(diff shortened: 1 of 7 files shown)


src/psygnal/_group_descriptor.py
```diff
@@ -268,7 +268,7 @@ def _setattr_and_emit_(self: object, name: str, value: Any) -> None:
 
             # don't emit if the signal doesn't exist or has no listeners
             signal: SignalInstance = group[name]
-            if len(signal) < 2 and not len(group._psygnal_relay):
+            if len(signal) < 1 and not len(group._psygnal_relay):
```
Czaki, line 271:
```suggestion
            if len(signal) < 1:
```

Czaki (commented):
Generally looks ok. One suggestion and one question about the possibility of small improvement.

src/psygnal/_group_descriptor.py
tlambert03: oh how nice! :)

src/psygnal/_group.py
```diff
@@ -63,10 +67,15 @@ def __init__(
         self._signals = signals
         self._sig_was_blocked: dict[str, bool] = {}
 
+    def _append_slot(self, slot: WeakCallback) -> None:
+        super()._append_slot(slot)
+        self._connect_relay()
```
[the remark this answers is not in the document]
tlambert03: added in 4fdec36

Czaki (approved):

PR #278, opened 2024-02-21
perf: Do not use reducer if there is no callback in `SingalInstance.resume`

Czaki:
This is followup to #277 

Remove the whole reducing operation if there is no callback.

### DIFF
--- a/src/psygnal/_signal.py
+++ b/src/psygnal/_signal.py
@@ -1113,6 +1113,10 @@ class SignalInstance:
         # EventedModel.update, it may be undefined (as seen in tests)
         if not getattr(self, "_args_queue", None):
             return
+        if len(self._slots) == 0:
+            self._args_queue = []
+            return
+
         if reducer is not None:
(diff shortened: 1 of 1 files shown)


src/psygnal/_signal.py
```diff
@@ -1127,6 +1127,10 @@ def resume(self, reducer: ReducerFunc | None = None, initial: Any = _NULL) -> No
         # EventedModel.update, it may be undefined (as seen in tests)
         if not getattr(self, "_args_queue", None):
             return
+        if len(self._slots) == 0:
+            self._args_queue = []
```
tlambert03, line 1131:
```suggestion
            self._args_queue.clear()
```

don't think object permanence really matters here, but we do `clear()` below, so might as well remain consistent here

Czaki:
Should I add test for it? (some mocking of reducer)

tlambert03:
works for me!

tlambert03:
actually... it's fine.  the logic is so simple, and it does get covered by tests already.   will merge.  thanks!

PR #279, opened 2024-02-23
feat: deduplicate events emision in nested properties

Czaki:
This PR migrates https://github.com/napari/napari/pull/6075 into psygnal to deduplicated event emission.

This will allow in future to drop the `napari.utils.events` module.

### DIFF
--- a/src/psygnal/_evented_model_utils.py
+++ b/src/psygnal/_evented_model_utils.py
@@ -8,13 +8,13 @@ if TYPE_CHECKING:
     import pydantic.version
 
     if pydantic.version.VERSION.startswith("2"):
-        from ._evented_model_v2 import EventedModel as EventedModelv1
+        from ._evented_model_v2 import EventedModel as EventedModelV1
     else:
-        from ._evented_model_v1 import EventedModel as EventedModelv2
+        from ._evented_model_v1 import EventedModel as EventedModelV2
 
 
 class ComparisonDelayer:
-    def __init__(self, target: EventedModelv1 | EventedModelv2) -> None:
+    def __init__(self, target: EventedModelV1 | EventedModelV2) -> None:
         self._target = target
 
     def __enter__(self) -> None:
--- a/src/psygnal/_evented_model_utils.py
+++ b/src/psygnal/_evented_model_utils.py
@@ -1,0 +1,30 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from types import TracebackType
+
+    import pydantic.version
+
+    if pydantic.version.VERSION.startswith("2"):
+        from ._evented_model_v2 import EventedModel as EventedModelv1
+    else:
+        from ._evented_model_v1 import EventedModel as EventedModelv2
+
+
+class ComparisonDelayer:
+    def __init__(self, target: EventedModelv1 | EventedModelv2) -> None:
+        self._target = target
+
+    def __enter__(self) -> None:
+        self._target._delay_check_semaphore += 1
+
+    def __exit__(
+        self,
+        exc_type: type[BaseException] | None,
+        exc_val: BaseException | None,
+        exc_tb: TracebackType | None,
+    ) -> None:
+        self._target._delay_check_semaphore -= 1
+        self._target._check_if_values_changed_and_emit_if_needed()
--- a/src/psygnal/_evented_model_v1.py
+++ b/src/psygnal/_evented_model_v1.py
@@ -15,6 +15,7 @@ from typing import (
     no_type_check,
 )
 
+from ._evented_model_utils import ComparisonDelayer
 from ._group import SignalGroup
 from ._group_descriptor import _check_field_equality, _pick_equality_operator
 from ._signal import Signal, SignalInstance
@@ -298,12 +299,12 @@ class EventedModel(BaseModel, metaclass=EventedMetaclass):
 
         class Config:
             allow_property_setters = True
-            property_dependencies = {"c": ["a", "b"]}
+            field_dependencies = {"c": ["a", "b"]}
 
     m = MyModel()
     assert m.c == [1, 1]
     m.events.c.connect(lambda v: print(f"c updated to {v}"))
-    m.a = 2  # prints 'c updated to [2, 1]'
+    m.a = 2 # prints 'c updated to [2, 1]'
     ```
 
     """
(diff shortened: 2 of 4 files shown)


tlambert03:
thanks @Czaki, would be happy to do what is needed here to let napari drop the internal event object.

looks like this has a significant negative impact on the time for setattr in EventedModel, but with and without connections

| Change   | Before [9cdce45f] <main>   | After [8c62e562]    |   Ratio | Benchmark (Parameter)                                           |
|----------|----------------------------|---------------------|---------|-----------------------------------------------------------------|
| +        | 815±4μs                    | 4.42±0.03ms         |    5.42 | benchmarks.EventedModelSuite.time_setattr_with_connections(100) |
| +        | 28.2±0.09μs                | 134±2μs             |    4.76 | benchmarks.EventedModelSuite.time_setattr_no_connections(10)    |
| +        | 276±2μs                    | 1.30±0.01ms         |    4.73 | benchmarks.EventedModelSuite.time_setattr_no_connections(100)   |
| +        | 210±2μs                    | 509±4μs             |    2.42 | benchmarks.EventedModelSuite.time_setattr_with_connections(10)  |
| +        | 68.2±0.4ms                 | 83.7±10ms           |    1.23 | benchmarks.EventedSetSuite.time_update_new(100000)              |


is the ComparisonDelayer fundamental to the deduplication? or can we look at adding that in a separate PR?

tlambert03:
(also, feel free to add benchmarks to the suite that show what is *gained* from this PR, so we can balance our understanding of what is lost)

tlambert03:
one more question: is the problem you were trying to solve in #6075 related only to events on computed properties?  If I run the test you've added here on main, it passes except for the very last line: `e_m.assert_called_once()` ... which is called twice.

If so, I'd like it if we could step back and discuss a broader question here first.  I have to say i think the whole ability to add property-setters to a pydantic model and to calculate events on them may well have been an over-implementation on my part to begin with.  It's a whole new thing that pydantic didn't support at the time, and now *does*  seem to be [supporting in v2](https://docs.pydantic.dev/2.0/usage/computed_fields/).  So, if anything, I think we should perhaps be deprecating and de-emphasizing this little feature in psygnal, instead of elaborating on it at the potential expense of performance for non-property-setting classes.

I do definitely want to support your usage of EventedModel over in napari, but i'd like to understand whether this is actually a good broader design decision here, or if we might just be "inheriting" complicated patterns from napari.

@andy-sweet, I'd be grateful for your input here as well

tlambert03:
> The general performance gain is to avoid heavy callbacks execution (for example, update gui).

Yep that makes sense, but as the benchmarks here are suggesting, it's possible that the performance gain for properties comes at the expense of performance of the (more important) regular fields.

So, we might need this to be an opt-in feature that only applies if someone is actively trying to use this relatively uncommon pattern

Czaki:
ok. The test was wrongly written. It checks multiple cases, which here are covered by other tests. So it could be simplified.

tests/test_evented_model.py
```diff
@@ -845,3 +848,50 @@ class Model(EventedModel):
         m.a = 3
     check_mock.assert_has_calls([call(Model, "a", 3, 1)])
     mock1.assert_called_once()
+
+
+def test_single_emit():
```
andy-sweet, line 853: This test name needs a more descriptive name and/or a docstring to explain what it's testing. Maybe considering moving it under `test_derived_events` as well, since it is related.

src/psygnal/_evented_model_v1.py
```diff
@@ -370,30 +407,43 @@ def __setattr__(self, name: str, value: Any) -> None:
...
-        with signal_instance.blocked():
-            self._super_setattr_(name, value)
+    def _setattr_default(self, name: str, value: Any) -> None:
+        """Will be overwritten by metaclass __new__."""
 
-        # if the value has changed we emit the event with new value
-        after = getattr(self, name)
-        if not _check_field_equality(type(self), name, after, before):
-            signal_instance.emit(after)  # emit event
+    def _setattr_with_dependants(self, name: str, value: Any) -> None:
```
andy-sweet, line 422: English spelling is a nightmare: dependents instead of dependants (this needs to be change in multiple places, so no code suggestion)

This follows US English spelling (which I think is consistent with the rest of the spelling in this repo) and is also consistent with the existing related `__field_dependents__`.

src/psygnal/_evented_model_utils.py
```diff
@@ -0,0 +1,30 @@
...
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from types import TracebackType
+
+    import pydantic.version
+
+    if pydantic.version.VERSION.startswith("2"):
+        from ._evented_model_v2 import EventedModel as EventedModelv1
```
andy-sweet, line 11: Nit: I would have expected `EventedModelV1` and `EventedModelV2` (i.e. with an upper case V, not lower case v).

tlambert03:
this is great! Thanks @Czaki!  I like the selection of setattr in `__new__`.  good solution.

I'm inclined to merge as is.  anything else you want to do here?

tlambert03 (approved):

src/psygnal/_evented_model_v1.py
```diff
@@ -343,15 +351,44 @@ def _super_setattr_(self, name: str, value: Any) -> None:
...
-            # fallback to default behavior
-            return self._super_setattr_(name, value)
+    def _check_if_values_changed_and_emit_if_needed(self) -> None:
+        """
+        Check if field values changed and emit events if needed.
 
+        The advantage of moving this to the end of all the modifications is
+        that comparisons will be performed only once for every potential change.
+        """
+        if self._delay_check_semaphore > 0 or len(self._changes_queue) == 0:
```
andy-sweet, line 361: Why aren't these covered?

andy-sweet (changes requested):
The only changes I'm requesting here are to fixing the annoying British/US English spelling of dependants/dependents and the name of the new test. Those are minor, but worth fixing in my opinion.

At a higher level, I agree with @tlambert03's sentiments that property setters were probably not a great idea and this PR makes `x.y = z` more difficult to understand as a developer. But `x.y = z` was already pretty difficult to understand and this does cover an extra case (even if it might be rare) and helps the napari migration. So I think it's worth it and I'll approve after the minor changes/suggestions described above.

In the long term, it would be nice to think about a way to implement all this desired behavior in a more understandable way. But let's leave that discussion elsewhere.

src/psygnal/_evented_model_v1.py
Czaki: I see 100% coverage.

tests/test_evented_model.py
Czaki: I think that edge case should have own test function

tests/test_evented_model.py
```diff
@@ -845,3 +848,52 @@ class Model(EventedModel):
         m.a = 3
     check_mock.assert_has_calls([call(Model, "a", 3, 1)])
     mock1.assert_called_once()
+
+
+def test_if_event_is_emited_only_once():
```
andy-sweet, line 853: Sorry, one last spelling nit.

```suggestion
def test_if_event_is_emitted_only_once():
```

andy-sweet (approved):
Looks good now except one new spelling mistake. Thanks for the quick updates!

src/psygnal/_evented_model_v1.py
tlambert03: yeah, that can happen when the tests aren't complete yet.  these are all covered. resolving

tlambert03:
awesome.  Thanks a lot both @Czaki for the nice enhancements, and @andy-sweet for the review.  Will merge now, and then cleanup a bit more in #280

PR #280, opened 2024-02-23
refactor: Unify pydantic evented model modules

tlambert03:
This PR unifies and deduplicates a lot of the pydantic evented model code that had a lot of copying to support pydantic v2 at first.


Czaki:
I think that #279 is ready for next review,

tlambert03:
just merged in #279  ... but won't merge this PR until that one is in

PR #281, opened 2024-02-24
fix: ensure proper order of signal emmision

Czaki:
I have noticed that If some callback triggers emission of an event during an existing loop, the events may come in incorrect order, leading to a corrupted state. 

I think that it is a bug, but I'm not sure if my approach to solve is correct.


src/psygnal/_signal.py
```diff
@@ -1034,14 +1040,25 @@ def __call__(
...
         with self._lock:
-            with Signal._emitting(self):
-                for caller in self._slots:
-                    try:
-                        caller.cb(args)
-                    except Exception as e:
-                        raise EmitLoopError(
-                            cb=caller, args=args, exc=e, signal=self
-                        ) from e
+            self._emit_queue.append(args)
```
tlambert03, line 1043: my first thought here is that the new `_emit_queue` seems awfully similar to the existing `_args_queue`, which already deals with the concept of queuing emission events.  So, perhaps we can unify those two mechanisms?  

I think that would mean that this line `self._emit_queue.append(args)` would move up to the `emit` method (i.e. it would pull the existing `self._args_queue.append(args)` call on line ~981 out of the the `if self._is_paused:` conditional)... and then `_run_emit_loop` would look more like `while self._args_queue: args = self._args_queue.pop()...`

tlambert03: and then, it also wouldn't need to be an exception to emit during emission... it would just assume the proper place at the end of the queue

Czaki: The exception is to allow preventing of object modification during signal emission. It may be useful for debug, but also for some fragile objects.  

In context of merge queues 

Lets see this code:

```python

s = EventedSet()

def callback1():
   if 1 not in s:
       s.add(1)

def callback2():
   if 5 not in s:
       s.update(range(5, 10)

s.events.items_changed.connect(callback1) 
s.events.items_changed.connect(callback2)

s.add(11)
```

What events you expect to be emitted

```
((11,),()),  ((1,),()), ((5,6,7,8,9),())
```
or 

```
((11,),()), ((1,5,6,7,8,9),())
```

As after merge queues, I do not see option to generate first one that is more accurate for me.

tlambert03: I think we're in agreement that the first emission pattern is the desired pattern (with (1,) separate from the 5-10). But I'm unsure what you mean by "merge queues".  I don't think that calling emit should merge the current emitted args with any existing ones, but rather that it should be added to a queue of emission.  Lemme see if I can put into code what I'm imagining.

tlambert03:
minor modification suggested in https://github.com/Czaki/psygnal/pull/1
it's not perfect (see note about recursion error), but curious if you have any thoughts on that approach vs this current one

tlambert03:
please merge main here and make sure all the tests we've discussed are added, then ping for review

tests/test_psygnal.py
```diff
@@ -988,3 +988,53 @@ def callback() -> None:
...
+
+    def callback(x):
+        if x < 10:
+            emitter.one_int.emit(10)
+
+    def callback2(x):
+        if x == 10:
+            emitter.one_int.emit(11)
+
+    def callback3(x):
```
tlambert03, line 1026: perhaps callback3 here is intended to simulate what the evented set would do? (without having to use evented set) ... but you could also add the following test for evented set just for the sake of thoroughness:

```python
def test_set_emission_order():
    s = EventedSet()

    def callback1():
        if 1 not in s:
            s.add(1)

    def callback2():
        if 5 not in s:
            s.update(range(5, 10))

    s.events.items_changed.connect(callback1)
    s.events.items_changed.connect(callback2)
    mock = Mock()
    s.events.items_changed.connect(mock)

    s.add(11)
    mock.assert_has_calls(
        [
            call((11,), ()),
            call((1,), ()),
            call((5, 6, 7, 8, 9), ()),
        ]
    )
```

Czaki:
@tlambert03 compiled test are ending with Segfault. Did you meet something like this before:

```
  File "/Users/runner/work/psygnal/psygnal/tests/test_psygnal.py", line 990 in test_recursion_error
```

tlambert03:
no, I haven't seen that before.  I can reproduce locally as well though (was just trying as you wrote), but not on main.  (btw, to test locally, you can run `make build`, and `make clean` to undo ...)

Czaki:
Ok. The problem is that hatch/mypyc performs compilation in tmp directory (unlike cython that places c/cpp files in the structure of packages. So the debugger cannot find the source files. 

I will try to dig, but maybe you know how to change this behavior or disable cleaning tmp 


```
#7  0x00007ffff508f242 in CPyDef__signal___SignalInstance____run_emit_loop ()
    at /tmp/tmpcocqfl4i/build/__native_ac51d50a4f4b6d748b8c.c:37311
37311   /tmp/tmpcocqfl4i/build/__native_ac51d50a4f4b6d748b8c.c: No such file or directory
```

tlambert03:
good sleuthing.  i also don't know, but happy to look into it if you can't find it easily

Czaki:
Im going to Warsaw Python meeting so will not be active today. Will sit to this at evening or toomorow

src/psygnal/_signal.py
```diff
@@ -944,22 +945,35 @@ def __call__(
...
+            if len(self._emit_queue) > 1:
+                return None
+            try:
+                self._run_emit_loop_inner()
+            finally:
+                self._emit_queue.clear()
 
         return None
```
tlambert03, line 978: here is a single _run_emit_loop that I feel reads much more clearly than the current two-function method that does funny stuff with `self._emit_queue = self._emit_queue[q_len:]`.  It also passes all the tests and works fine with mypyc.

```python
    def _run_emit_loop(self, args: tuple[Any, ...]) -> None:
        # allow receiver to query sender with Signal.current_emitter()
        with self._lock:
            self._emit_queue.append(args)
            if Signal.current_emitter() is self:
                return
            try:
                with Signal._emitting(self):
                    count = 0
                    while self._emit_queue:
                        args = self._emit_queue.pop(0)
                        for caller in self._slots:
                            caller.cb(args)
                        count += 1
                        if count > sys.getrecursionlimit():
                            raise RecursionError
            except RecursionError as e:
                raise RecursionError(f"RecursionError in signal {self.name!r}") from e
            except Exception as e:
                raise EmitLoopError(cb=caller, args=args, exc=e, signal=self) from e
```

 `self._emit_queue` can easily be made a deque.  If you don't like this, or if you have concerns about thread safety.  Can you perhaps provide a test that fails with the code above, but not with your code?  Because otherwise, i find it much easier to reason about.

src/psygnal/_signal.py
```diff
@@ -947,19 +950,29 @@ def __call__(
...
+                return None
+            try:
+                with Signal._emitting(self):
+                    i = 0
+                    while i < len(self._emit_queue):
+                        args = self._emit_queue[i]
+                        for caller in self._slots:
+                            caller.cb(args)
+                            if len(self._emit_queue) > RECURSION_LIMIT:
+                                raise RecursionError
```
Czaki, line 965: This simulated stack is 2-3x times bigger than the original one, but maybe it not a problem.

src/psygnal/_signal.py
```diff
@@ -947,19 +950,29 @@ def __call__(
...
+                raise RecursionError(
+                    f"RecursionError in {caller.slot_repr()} when "
+                    f"emitting signal {self.name!r} with args {args}"
+                ) from e
+            except Exception as e:
+                raise EmitLoopError(cb=caller, args=args, exc=e, signal=self) from e
+            finally:
+                self._emit_queue.clear()
 
         return None
```
tlambert03, line 977: Works for me!  Thanks.


I'm not worried about the very small benchmark decreases on emit to setattr and setitem for large numbers of callbacks.

Thanks for working on this

tlambert03 (approved):

PR #285, opened 2024-02-27
feat: add priority to connect, to control callback order

tlambert03:
This adds the ability to control the order in which callbacks are called at connection time.  Higher priority connections are called first, and negative priorities are allowed.  with default 0

```python
sig = SignalInstance()

calls = []

sig.connect(lambda: calls.append(1))
sig.connect(lambda: calls.append(2), priority=5)
sig.connect(lambda: calls.append(3), priority=-5)
sig.connect(lambda: calls.append(4), priority=10)

sig.emit()
assert calls == [4, 2, 1, 3]
```

### DIFF
--- a/src/psygnal/_group.py
+++ b/src/psygnal/_group.py
@@ -440,6 +440,7 @@ class SignalGroup:
         unique: bool | str = ...,
         max_args: int | None = None,
         on_ref_error: RefErrorChoice = ...,
+        priority: int = ...,
     ) -> Callable[[F], F]: ...
 
     @overload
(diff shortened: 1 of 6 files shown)


tlambert03:
@Czaki, let me know if you have any feelings about this feature or it's implementation.
This is similar to `position` in vispy's events... but without the whole `before`/`after` callbacks.  The connector simply gives a priority number by which to order the callbacks; with zero being the default, and higher priority callbacks getting placed before lower priority ones.

PR #286, opened 2024-02-29
refactor: Add back SignalGroup methods

tlambert03:
just running tests here, then will likely force push this to #274


tlambert03:
ok @getzze, I think we're there.  I pulled it out into a new PR so that I could focus on a couple things.  I'd be happy to either force push this branch onto yours, or just add you as an author to the squash-merge here, whichever you prefer.

and, if you have time, I'd appreciate your eyes on this

src/psygnal/_group.py
```diff
@@ -251,53 +256,74 @@ class MySignals(SignalGroup):
...
-        if not hasattr(cls, "_psygnal_signals"):  # pragma: no cover
+        if not hasattr(cls, "_psygnal_signals"):
             raise TypeError(
                 "Cannot instantiate `SignalGroup` directly.  Use a subclass instead."
             )
 
         self._psygnal_instances = {
-            name: sig.__get__(self, cls) for name, sig in cls._psygnal_signals.items()
+            name: (
+                sig._create_signal_instance(self)
```
tlambert03, line 273: :eyes: @getzze 

here was my solution for using the standard Signal.__get__ descriptor protocol for non-conflicting names, and a (non-Signal-cached) local instance for conflicting names.

getzze:
I was checking the restrictions for using `mypyc` and some parts are not compatible with it, namely that you cannot [delete](https://mypyc.readthedocs.io/en/latest/native_classes.html#deleting-attributes) attributes from classes and classes [don't have `__dict__`](https://mypyc.readthedocs.io/en/latest/native_classes.html#other-properties).

Then maybe the implementation can be simplified by creating a factory function to create `SignalGroup` subclasses to be used in `SignalGroupDescriptor`, allowing passing a keyword argument with conflicting signals on subclassing `SignalGroup`, then just raise an error in `SignalGroup` creation if there is an attribute name conflict.

Also, we should call `__set_name__` on conflicting `Signal`s otherwise they have an empty name.

```python
def signal_group_forbidden_attr(name: str):
    return name.startswith("_psygnal")

def signal_group_reserved_attr(name: str):
    reserved = set(dir(SignalGroup))
    return name in reserved or name.startswith("psygnal")

def factory_signal_group(name: str, signals: dict[str, Signal], strict: bool = False) -> SignalGroup:
    conflicts = {}
    for key, sig in signals.copy().items():
        if signal_group_forbidden_attr(key):
            # forbidden
            del signals[key]
            continue

        if signal_group_reserved_attr(key):
            conflicts[key] = sig
            del signals[key]

        # Special SignalInstance
        sig._signal_instance_class = _DataclassFieldSignalInstance

    if conflicts:
        Names = "Names" if len(conflicts) > 1 else "Name"
        Are = "are" if len(conflicts) > 1 else "is"
        warnings.warn(
            f"{Names} {sorted(conflicts)!r} {Are} reserved. You cannot use these "
            "names to access SignalInstances as attributes on a SignalGroup. (You "
            "may still access them as keys to __getitem__: `group['name']`).",
            UserWarning,
            stacklevel=2,
        )

    print(f"{signals=}, {conflicts=}")
    return type(name, (SignalGroup,), signals, reserved_signals=conflicts)


    def __init_subclass__(
        cls,
        strict: bool = False,
        reserved_signals: Mapping[str, Signal] = {},
    ) -> None:
        """Collects all Signal instances on the class under `cls._psygnal_signals`."""
        # Collect Signals and remove from class attributes
        # Use dir(cls) instead of cls.__dict__ to get attributes from super()
        forbidden = {
            k for k in dir(cls)
            if isinstance(getattr(cls, k), Signal) and (signal_group_forbidden_attr(k) or signal_group_reserved_attr(k))
        }

        if forbidden:
            raise TypeError(
                f"SignalGroup subclass cannot have attributes starting with '_psygnal'."
                f" Found: {forbidden}"
            )

        _psygnal_signals = {}
        for k in dir(cls):
            val = getattr(cls, k, None)
            if isinstance(val, Signal):
                _psygnal_signals[k] = val

        for k, sig in reserved_signals.items():
            sig.__set_name__(cls, k)
        cls._psygnal_name_conflicts = reserved_signals

        # Collect the Signals also from super-class
        # When subclassing, the Signals have been removed from the attributes,
        # look for cls._psygnal_signals also
        cls._psygnal_signals = {
            **getattr(cls, "_psygnal_signals", {}),
            **_psygnal_signals,
            **reserved_signals,
        }

        cls._psygnal_uniform = _is_uniform(cls._psygnal_signals.values())
        if strict and not cls._psygnal_uniform:
            raise TypeError(
            )
        super().__init_subclass__()


```

tlambert03:
Thanks!

> I was checking the restrictions for using mypyc and some parts are not compatible with it, namely that you cannot delete attributes from classes and classes don't have __dict__.

Can you make a test that fails according to what you think the issue would be?  Because we do test conflicts on both compiled and uncompiled code here... and it seems to be working fine 


> Also, we should call __set_name__ on conflicting Signals otherwise they have an empty name.

👍

getzze:
However, I have doubts if it's a good idea to have real and fake `Signal` attributes in `SignalGroup`.

An example of unexpected behavior:
```python
class MyGroup(SignalGroup):
    sig: Signal[int]
    connect: Signal[bool]

hasattr(MyGroup, "sig")  # True
hasattr(MyGroup, "connect")  # True
delattr(MyGroup, "sig")  # success
delattr(MyGroup, "connect")  # AttributeError, trying to delete the method on `SignalGroup`
```


To go back to your example, why would somebody want to do `hasattr(MyGroup, "sig")`? If it's to check if the **signal** `sig` is in the group, then it's not a safe method, as conflicting signal names would point to the method anyway. One would need a dedicated function or classmethod to check if the `Signal` is in the class instead.

It seems safer to me to document well that `SignalGroup` is a special class, first that it has to be subclassed, second you can define the `Signal`s attached to the group using syntactic sugar, although the `Signal`s end up not being attributes of the class. So the primary way of subclassing `SignalGroup` would be with a factory function, the above example would just be syntactic sugar.

getzze:
> Can you make a test that fails according to what you think the issue would be? Because we do test conflicts on both compiled and uncompiled code here... and it seems to be working fine

I didn't test very well before posting, actually this code is not even working correctly...

I didn't know it was tested on compiled code also and that it worked fine, then forget about it then. What I understood from `mypyc` documentation is that these were the restrictions of native classes, but then `SignalGroup` is probably not a native class.

getzze:
> However, I have doubts if it's a good idea to have real and fake `Signal` attributes in `SignalGroup`.
> 
> An example of unexpected behavior:
> 
> ```python
> class MyGroup(SignalGroup):
>     sig = Signal[int]
>     connect = Signal[bool]
> 
> hasattr(MyGroup, "sig")  # True
> hasattr(MyGroup, "connect")  # True
> delattr(MyGroup, "sig")  # success
> delattr(MyGroup, "connect")  # AttributeError, trying to delete the method on `SignalGroup`
> ```
> 
> To go back to your example, why would somebody want to do `hasattr(MyGroup, "sig")`? If it's to check if the **signal** `sig` is in the group, then it's not a safe method, as conflicting signal names would point to the method anyway. One would need a dedicated function or classmethod to check if the `Signal` is in the class instead.
> 
> It seems safer to me to document well that `SignalGroup` is a special class, first that it has to be subclassed, second you can define the `Signal`s attached to the group using syntactic sugar, although the `Signal`s end up not being attributes of the class. So the primary way of subclassing `SignalGroup` would be with a factory function, the above example would just be syntactic sugar.

Quoting myself here, but with the factory function this unexpected behavior would go away, so I solved my own problem :laughing:

tlambert03:
The main issue I have with the factory function is that it's a solution only for evented dataclasses, but not for directly created SignalGroups subclasses.  So, i'd like to focus on the current solution and determine if there are any tests we can make that show undesirable behavior.

```python
delattr(MyGroup, "connect")  # AttributeError, trying to delete the method on `SignalGroup`
```

to be honest i'm not worried about this example at all.  It's likely to be a very rare attempt (deleting a signal only in the case of a conflict).  The user would have already received a warning that something funny was up with `connect`, and it's actually the correct exception to raise at that point.

Were there any other issues you can demonstrate with the current pattern?

getzze:
```python
class MyGroup(SignalGroup):
    sig = Signal(int)
    connect = Signal(bool)

g = MyGroup()
```
I tried making a copy and deepcopy of `MyGroup`, a copy and deepcopy of `g`, and `sig` and `connect` behave identically. I don't know what else to try, maybe pickling?

tlambert03:
> I tried making a copy and deepcopy of MyGroup, a copy and deepcopy of g, and sig and connect behave identically. I don't know what else to try, maybe pickling?

can you give me an `assert` line to show what you expect to happen?

getzze:
I run `test_group.test_group_deepcopy` with `MyGroup` with `sig` and `connect` signals. I'll paste the code.

tlambert03:
that's enough info, thanks.  I'll parametrize that test with a group with conflicts and work on it

tlambert03:
added a test: https://github.com/pyapp-kit/psygnal/pull/286/commits/5b1f7014cfdae95d0c9ef78be84e0423d2024182

it still seems to be working though... so you might indeed need to paste some code if you think you found a case that raises an unexpected error

tlambert03:
also added tests to make sure that the __set_name__ is called on both conflicting and non-conflicting names on SignalGroups and all their subclasses.  They were, and we don't need to call it explicitly

getzze:
> it still seems to be working though... so you might indeed need to paste some code if you think you found a case that raises an unexpected error

No, no, I didn't find any error, all working great!

tlambert03:
Oh good!  Ok, I'll merge then so I can cut a new release without the deprecations, and so you can finally actually use the original previous value emission feature that you originally added here ;)

PR #289, opened 2024-03-06
Uuse self._psygnal_instances consistently

getzze:


tlambert03:
looks good to me!  thanks @getzze.  
side question: I notice you mark most of your PRs as draft, and don't seem to mark them as ready for review.  Should I assume that they're ready to go if all tests are passing and you haven't touched it for a few days?

getzze:
I'm not 100% when to use draft or not, but this PR is not a draft indeed :)

PR #290, opened 2024-03-06
build arm64 on macos-14

tlambert03:
just trying

### DIFF
--- a/.github/workflows/pypi.yml
+++ b/.github/workflows/pypi.yml
@@ -4,6 +4,8 @@ on:
   push:
     branches: [main]
     tags: [v*]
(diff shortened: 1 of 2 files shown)


tlambert03:
no good:

> Warning: While cibuildwheel can build CPython 3.8 universal2/arm64 wheels, we cannot test the arm64 part of them, even when running on an Apple Silicon machine. This is because we use the x86_64 installer of CPython 3.8. See the discussion in https://github.com/pypa/cibuildwheel/pull/1169 for the details. To silence this warning, set `CIBW_TEST_SKIP: "cp38-macosx_*:arm64"`.

PR #291, opened 2024-03-06
SignalGroupDescriptor collects fields with a SignalGroup subclass

getzze:
To be implemented `_build_dataclass_signal_group` cannot be cached with `lru_cache`, so the result is cached in the `SignalGroupDescriptor` instead.

PR extracted from #265


src/psygnal/_group_descriptor.py
```diff
@@ -162,7 +176,9 @@ def _build_dataclass_signal_group(
         # patch in our custom SignalInstance class with maxargs=1 on connect_setattr
         sig._signal_instance_class = _DataclassFieldSignalInstance
 
-    return type(f"{cls.__name__}SignalGroup", (SignalGroup,), signals)
+    # Create `signal_group_class` subclass with the attached signals
+    group_name = f"{cls.__name__}_{signal_group_class.__name__}"
```
tlambert03, line 180:
```suggestion
    group_name = f"{cls.__name__}{signal_group_class.__name__}"
```

should be camel case right?

tlambert03 (approved):
very nice. Good to go by me, one very minor comment on camel case naming of the created group class

src/psygnal/_group_descriptor.py
getzze: Yes!

getzze:
The PR changes the previous default behavior when a subclass of `SignalGroup` is provided (now it collects the fields). But this default behavior was not clearly documented so I believe it's fine.

Also, `signal_group_class=SignalGroup` and `collect_fields=False` fails without a proper error message and only when accessing `foo.events`. I was not sure what was best, or raising an error at `SignalGroupDescriptor.__init__` or fallback to `collect_fields=True` with a warning.

tlambert03:
> But this default behavior was not clearly documented so I believe it's fine.

i agree. and i think it's a good improvement thanks.

> Also, signal_group_class=SignalGroup and collect_fields=False fails without a proper error message and only when accessing foo.events. I was not sure what was best, or raising an error at SignalGroupDescriptor.__init__ or fallback to collect_fields=True with a warning.

ah yeah, that's probably a pretty easy error to run into, since it only requires setting `collect_fields=False` (as the other is the default).  So I agree, add an exception in the `__init__` if `collect_fields is False and signal_group_class is SignalGroup`

one more thing:

let's make `SignalGroupDescriptor` a generic now: `SignalGroupDescriptor(Generic[GroupType])`

I'll make some comments inline

tlambert03:
actually, it's a bit hard to do inline since you haven't touched all the lines needed here.  i'll make a PR to your PR

getzze:
ok, I will add the early error then.

tlambert03:
hang on just a minute... PR incoming

tlambert03:
https://github.com/getzze/psygnal/pull/1

tlambert03:
(sorry if that gave merge conflicts) :)

tlambert03:
ah whoops, looks like I forgot to add `from __future__ import annotations` in test_group_descriptor

tlambert03:
arggg... pydantic errors in the test_group_descriptor.  I had you add __future__.annotations thinking there wasn't any runtime typing used in that test module, but it looks like there was after all.  So i guess we'll need to remove that import at the top, and use `Optional[Type[SignalGroup]]` instead of `type[SignalGroup] | None`

getzze:
ok, I'll make the changes.

getzze:
mmmh there are other problems with types and coverage. Maybe I'll let you deal with them :), you can modify this PR directly if you don't mind.

tlambert03:
>  other problems with types and coverage. Maybe I'll let you deal with them :)

will do!  was just noticing that.  Going to add a `# pragma: no cover` and be done with it

tlambert03:
I reverted the generic stuff, and will deal with it in another PR.  The reason is that, as implemented, it makes it unnecessarily ugly to know that the `__get__` method of the descriptor returns a subclass of SignalGroup:

```python
@dataclass
class T:
    x: int = 0
    e: ClassVar[SignalGroupDescriptor] = SignalGroupDescriptor()

t = T()
reveal_type(T.e)    # N: Revealed type is "psygnal._group_descriptor.SignalGroupDescriptor[Any]" ... fine
reveal_type(t.e)    # N: Revealed type is "Any" .... BAD
```

PR #293, opened 2024-03-08
Add recursion_mode ('immediate' or 'deferred') to Signal and SignalInstance

### LINKED ISSUE #292: RecursionError with v0.10.0
* psygnal version: 0.10.0
* Python version: 3.11.4
* Operating System: Windows 11

### Description

I found in a specific case, psygnal<0.10 works fine but recursion error occurs in the emission loop with 0.10.0.

### What I Did

This is a simple example that class `A` has an evented property `value` and class `Linker` can link the values of different `A` instances.

``` python
from psygnal import SignalGroup, Signal

class Events(SignalGroup):
    value = Signal(int)

class A:
    def __init__(self):
        self.events = Events()
        self._value = 0
    
    @property
    def value(self):
        return self._value
    
    @value.setter
    def value(self, val: int):
        self._value = val
        self.events.value.emit(val)

class Linker:
    def __init__(self, objects: list[A]):
        self._objs = objects
        for obj in objects:
            obj.events.value.connect(self.set_values)
        self._updating = False
    
    def set_values(self, val: int):
        if self._updating:
            return  # avoid recursion error
        self._updating = True
        try:
            for obj in self._objs:
                obj.value = val
        finally:
            self._updating = False
        
a0 = A()
a1 = A()

linker = Linker([a0, a1])

a0.value = -1
assert a1.value == -1
```

```
RecursionError                            Traceback (most recent call last)
File src\psygnal\_signal.py:1014, in _run_emit_loop()

RecursionError: 

During handling of the above exception, another exception occurred:

RecursionError                            Traceback (most recent call last)
Cell In[1], line 42
     38 a1 = A()
     40 linker = Linker([a0, a1])
---> 42 a0.value = -1
     43 assert a1.value == -1

Cell In[1], line 18, in A.value(self, val)
     15 @value.setter
     16 def value(self, val: int):
     17     self._value = val
---> 18     self.events.value.emit(val)

File src\psygnal\_signal.py:986, in emit()

File src\psygnal\_signal.py:1017, in _run_emit_loop()

RecursionError: RecursionError in __main__.Linker.set_values when emitting signal 'value' with args (-1,)
```

tlambert03:
fixes #292 

This adds the possibility to control the order callbacks during recursive emission events (when one of the callbacks itself emits the same event):

- `"immediate"`: Nested emission events immediately begin a deeper emission loop:
calling all callbacks with the arguments of the nested emitter before
returning to continue the original emission loop.  This is the default
behavior.
- `"deferred"`: Nested emission events are deferred until the current emission
loop is complete.  This is the default behavior.

### DIFF
--- a/src/psygnal/_signal.py
+++ b/src/psygnal/_signal.py
@@ -310,6 +310,16 @@ class SignalInstance:
         Whether to check the callback parameter types against `signature` when
         connecting a new callback. This can also be provided at connection time using
         `.connect(..., check_types=True)`. By default, False.
+    recursion_mode : Literal["immediate", "deferred"]
+        How to handle recursive emissions, when this signal is emitted by a callback
+        while this signal is being emitted.
+
+        - "immediate": Nested emission events immediately begin a deeper emission loop:
+          calling all callbacks with the arguments of the nested emitter before
+          returning to continue the original emission loop.  This is the default
+          behavior.
+        - "deferred": Nested emission events are deferred until the current emission
+          loop is complete.  This is the default behavior.
 
     Raises
     ------
@@ -330,11 +340,8 @@ class SignalInstance:
         name: str | None = None,
         check_nargs_on_connect: bool = True,
         check_types_on_connect: bool = False,
+        recursion_mode: Literal["immediate", "deferred"] = "deferred",
     ) -> None:
-        self._name = name
-        self._instance: Callable = self._instance_ref(instance)
-        self._args_queue: list[tuple] = []  # filled when paused
-
         if isinstance(signature, (list, tuple)):
             signature = _build_signature(*signature)
         elif not isinstance(signature, Signature):  # pragma: no cover
@@ -343,6 +350,15 @@ class SignalInstance:
                 "instance of `inspect.Signature`"
             )
 
+        if recursion_mode not in ("immediate", "deferred"):
+            raise ValueError(
+                "recursion_mode must be one of 'immediate' or 'deferred', not "
+                f"{recursion_mode!r}"
+            )
+
+        self._name = name
+        self._instance: Callable = self._instance_ref(instance)
+        self._args_queue: list[tuple] = []  # filled when paused
         self._signature = signature
         self._check_nargs_on_connect = check_nargs_on_connect
         self._check_types_on_connect = check_types_on_connect
@@ -351,6 +367,7 @@ class SignalInstance:
         self._is_paused: bool = False
         self._lock = threading.RLock()
         self._emit_queue: list[tuple] = []
+        self._recursion_mode = recursion_mode
         # whether any slots in self._slots have a priority other than 0
         self._priority_in_use = False
 
@@ -984,7 +1001,6 @@ class SignalInstance:
             SignalInstance._debug_hook(EmissionInfo(self, args))
 
         self._run_emit_loop(args)
-        return None
 
     def __call__(
         self, *args: Any, check_nargs: bool = False, check_types: bool = False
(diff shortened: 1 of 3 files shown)


src/psygnal/_signal.py
```diff
@@ -996,7 +1031,32 @@ def __call__(
...
+                        caller.cb(args)
+                    except RecursionError as e:
+                        raise RecursionError(
+                            f"RecursionError in {caller.slot_repr()} when "
+                            f"emitting signal {self.name!r} with args {args}"
+                        ) from e
+                    except Exception as e:
+                        raise EmitLoopError(
+                            cb=caller, args=args, exc=e, signal=self
+                        ) from e
```
Czaki, line 1055: I am not convinced that chain of few hundreds exceptions (3k recursion limit) is a good idea. Maybe information should be added once.

src/psygnal/_signal.py
```diff
@@ -310,6 +323,16 @@ class Emitter:
...
+    recursion_mode : Literal["immediate", "deferred"]
+        How to handle recursive emissions, when this signal is emitted by a callback
+        while this signal is being emitted.
+
+        - "immediate": Nested emission events immediately begin a deeper emission loop:
+          calling all callbacks with the arguments of the nested emitter before
+          returning to continue the original emission loop.  This is the default
+          behavior.
+        - "deferred": Nested emission events are deferred until the current emission
+          loop is complete.  This is the default behavior.
```
Czaki, line 335:
```suggestion
        - "deferred": Nested emission events are deferred until the current emission
          loop is complete.
```

src/psygnal/_signal.py
```diff
@@ -107,6 +107,16 @@ class attribute that is bound to the signal will be used. default None
...
```
Czaki, line 119:
```suggestion
        - "deferred": Nested emission events are deferred until the current emission
          loop is complete.
```

Czaki (commented):
I prefer to use flag, instead of a boolean variable, as it will allow extending in future.  

I would like to see in this PR update of EventedModel to use dereferred mode by default.

I thinks that it will be nice to document somewhere why dereferred mode is better in some cases.

src/psygnal/_signal.py
tlambert03: i changed the logic a bit to wrap the inner bit (calling `caller.cb`) to unify the exception handling between the two modes.  I'm going to follow up on that in a follow up PR, but for this PR, I think they should now both be behaving like the loop you defined in #281

tlambert03:
The failing benchmarks represent a slight increase in the time to create a signal instance:


| Change   | Before [f4ac7c3f] <v0.10.0^0>   | After [b17be85c]    |   Ratio | Benchmark (Parameter)                                      |
|----------|---------------------------------|---------------------|---------|------------------------------------------------------------|
| +        | 477±3ns                         | 833±3ns             |    1.74 | benchmarks.CreateSuite.time_create_signal_instance         |
| +        | 4.80±0.05μs                     | 5.87±0.08μs         |    1.22 | benchmarks.EventedSetSuite.time_create_set(10)             |
| +        | 4.85±0.03μs                     | 5.92±0.02μs         |    1.22 | benchmarks.EventedSetWithCallbackSuite.time_create_set(10) |

which is not something I'm worried about here given that it's on the order of ns.

@Czaki, this is ready for final review.  I'll wait for your review on this one before merging. And I intend to follow up in another PR reviewing the exception handling logic slightly

Czaki (commented):
In general, looks ok, with one remark.

src/psygnal/_evented_model.py
```diff
@@ -213,10 +215,31 @@ def __new__(
         model_fields = _get_fields(cls)
         model_config = _get_config(cls)
 
+        recursion_cfg = model_config.get(RECURSION_MODE, {})
+        default_recursion: RecursionMode = DEFAULT_RECURSION_MODE
```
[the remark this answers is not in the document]
tlambert03: ok, made that change.

will add more documentation in a followup PR

PR #296, opened 2024-03-11
feat: Rename `recursion_mode` to `reemission`.  Rename `deferred` to `queued`.  Add `latest-only` mode.

tlambert03:
This PR renames the parameter controlling how callbacks are (re-)invoked when one of the callbacks itself emits the signal currently being emitted.  From the new docstrings:

```
reemission : Literal["immediate", "queued", "latest-only"] | None
    Determines the order and manner in which connected callbacks are invoked when a
    callback re-emits a signal. Default is `"immediate"`.
```

  * `"immediate"`: Signals emitted by callbacks are immediately processed in a
      deeper emission loop, before returning to process signals emitted at the
      current level (after all callbacks in the deeper level have been called).

  * `"queued"`: Signals emitted by callbacks are enqueued for emission after the
      current level of emission is complete. This ensures *all* connected
      callbacks are called with the first emitted value, before *any* of them are
      called with values emitted while calling callbacks.

  * `"latest-only"`: Signals emitted by callbacks are immediately processed in a
      deeper emission loop, and remaining callbacks in the current level are never
      called with the original value.

@Czaki, I was thinking this might actually be the desired behavior for evented models, since it not only makes sure that the current value is always up to date, it also prevents emitting things that are just about to become irrelevant

### DIFF
--- a/src/psygnal/_signal.py
+++ b/src/psygnal/_signal.py
@@ -1,6 +1,6 @@
 """The main Signal class and SignalInstance class.
 
-A note on "emission_strategy" in Signal and SignalInstances.  Since it can be a little
+A note on "reemission" in Signal and SignalInstances.  Since it can be a little
 confusing, take the following example of a Signal that emits an integer.  We'll connect
 three callbacks to it, two of which re-emit the same signal with a different value
 (this is the tricky condition we're trying to handle here)
@@ -52,7 +52,7 @@ calling cb2 with: 3
 calling cb3 with: 3
 ```
 
-with `emission_mode='nested'` signal emission from within
+with `emission_mode='immediate'` signal emission from within
 any of the callbacks immediately goes into the next deeper nested loop of emission
 events, before returning back to the original loop to call the later callbacks with the
 original value.
@@ -119,7 +119,7 @@ So: with EventedModel.setattr you can easily end up with some complicated recurs
 behavior if you connect an on-change callback that also sets the value of the model.
 In this case `emit_mode='depth-first'` is probably the most appropriate, as it will
 prevent the callback from being called with the original (now-stale) value.  But
-one can conceive of other scenarios where `emit_mode='nested'` or
+one can conceive of other scenarios where `emit_mode='immediate'` or
 `emit_mode='sequential'` might be more appropriate.
 
 """
@@ -133,6 +133,7 @@ import warnings
 import weakref
 from collections import deque
 from contextlib import contextmanager, suppress
+from enum import Enum
 from functools import lru_cache, partial, reduce
 from inspect import Parameter, Signature, isclass
 from typing import (
@@ -167,6 +168,8 @@ from ._weak_callback import (
 )
 
 if TYPE_CHECKING:
+    from typing import Self
+
     from ._group import EmissionInfo
     from ._weak_callback import RefErrorChoice
 
@@ -179,13 +182,38 @@ if TYPE_CHECKING:
 
 
 __all__ = ["Signal", "SignalInstance", "_compiled"]
-EmissionStrategy = Literal["sequential", "nested", "depth-first"]
 
 _NULL = object()
 F = TypeVar("F", bound=Callable)
 RECURSION_LIMIT = sys.getrecursionlimit()
-DEFAULT_EMISSION_STRATEGY: EmissionStrategy = "nested"
-EMISSION_STRATEGIES: set[EmissionStrategy] = set(EmissionStrategy.__args__)  # type: ignore
+
+
+ReemissionVal = Literal["immediate", "queued", "latest-only"]
+
+
+class ReemissionMode(str, Enum):
+    """Enumeration of reemission strategies."""
+
+    IMMEDIATE = "immediate"
+    QUEUED = "queued"
+    LATEST = "latest-only"
+
+    @classmethod
+    def cast(cls, value: str | ReemissionMode) -> Self:
+        if isinstance(value, str):
+            value = value.lower().replace("_", "-")
+
+        try:
+            return cls(value)
+        except ValueError:
+            raise ValueError(
+                "Invalid reemission value. Must be one of "
+                f"{', '.join(repr(x.value) for x in cls)}. Not {value!r}"
+            ) from None
+
+
+DEFAULT_REEMISSION: ReemissionVal = ReemissionMode.IMMEDIATE.value
+VALID_REEMISSION: set[ReemissionVal] = {R.value for R in ReemissionMode}
 
 
 class Signal:
@@ -239,23 +267,22 @@ class Signal:
         Whether to check the callback parameter types against `signature` when
         connecting a new callback. This can also be provided at connection time using
         `.connect(..., check_types=True)`. By default, `False`.
-    emission_strategy : Literal['nested', 'sequential', 'depth-first'] | None
+    reemission : Literal["immediate", "queued", "latest-only"] | None
         Determines the order and manner in which connected callbacks are invoked when a
-        callback re-emits a signal. Default is `"nested"`.
+        callback re-emits a signal. Default is `"immediate"`.
 
-        * `"nested"`: Signals emitted by callbacks are immediately processed in
-          a deeper emission loop, before returning to process signals emitted at the
-          current level.
+        * `"immediate"`: Signals emitted by callbacks are immediately processed in a
+            deeper emission loop, before returning to process signals emitted at the
+            current level (after all callbacks in the deeper level have been called).
 
-        * `"sequential"`: Ensures *all* connected callbacks are called with the first
-          emitted value, before *any* of them are called with values emitted while
-          calling callbacks. Signals emitted by callbacks are enqueued for emission
-          after the current level of emission is complete.
+        * `"queued"`: Signals emitted by callbacks are enqueued for emission after the
+            current level of emission is complete. This ensures *all* connected
+            callbacks are called with the first emitted value, before *any* of them are
+            called with values emitted while calling callbacks.
 
-        * `"depth-first"`: This is similar to `"nested"`, but control never returns to
-          the previous level of emission. When a callback emits a signal, a deeper
-          level of emission is started, and the current level is discontinued.
-          This strategy follows a depth-first execution without backtracking.
+        * `"latest-only"`: Signals emitted by callbacks are immediately processed in a
+            deeper emission loop, and remaining callbacks in the current level are never
+            called with the original value.
     """
 
     # _signature: Signature  # callback signature for this signal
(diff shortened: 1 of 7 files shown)


Czaki (commented):
Looks great. Few minor remarks

tlambert03:
side question:  Do you think you would probably use this mode over 'deferred'?

tests/test_psygnal.py
```diff
@@ -1019,13 +1020,18 @@ def cb3(value: int) -> None:
         # nested emission events occur immediately,
         # before proceeding to the next callback
         assert a == [1, 2, 20, 3, 30, 300, 200, 10, 100]
+    elif recursion == "immediate-drop":
+        # nested emission events occur immediately,
+        # before proceeding to the next callback
+        # AND any remaining emissions in the current loop level are cancelled
+        assert a == [1, 2, 20, 3, 30, 300]
```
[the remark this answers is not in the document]
tlambert03: eh, i thought it was kinda helpful to have the explanation comment right next to the expectation.  I could put those comments it up in the parameters, but then if you hit an error, you have to go up and find the comment associated with the error.  if we get a ton more modes, we can add the expectation to the parameter, but for now I find this easier to read

src/psygnal/_signal.py
```diff
@@ -1053,15 +1072,28 @@ def _run_emit_loop(self, args: tuple[Any, ...]) -> None:
...
                 ) from e
             finally:
+                self._recursion_depth -= 1
+                # we're back to the root level of the emit loop, reset max_depth
+                if self._recursion_depth <= 0:
+                    self._max_recursion_depth = 0
                 self._emit_queue.clear()
 
     def _run_emit_loop_immediate(self) -> None:
         self._args = args = self._emit_queue.popleft()
```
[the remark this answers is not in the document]
tlambert03: that is only there for raise `EmitLoopError(cb=self._caller, args=self._args)` ... it was a small change required to let us unify the inner loops, and keep all of the exception handling in the main `_run_emit_loop` logic.  I have a local branch that is playing with getting rid of this altogether, and having the EmitLoopError itself do stack inspection to grab this information... but that is difficult in the mypyc-compiled version.  So, stay tuned for that change :)

tlambert03:
oh and also: do you like the name `immediate-drop`? or can you think of a better one?

Czaki:
> side question: Do you think you would probably use this mode over 'deferred' for cases like EventedModel when the primary goal of the emitter is to emit the current value?

Yes. I think it will be better, as in most use cases I need to know only the last value. 



> oh and also: do you like the name `immediate-drop`? or can you think of a better one?

I do not have a better name, but I'm bad at naming. Maybe @jni could help.

src/psygnal/_signal.py
```diff
@@ -118,6 +121,10 @@ class attribute that is bound to the signal will be used. default None
           calling all callbacks with the arguments of the nested emitter before
           returning to continue the original emission loop.  This is the default
           behavior.
+        - "immediate-drop": Nested emission events immediately begin a deeper emission
+            loop, and the original emission loop is dis-continued.  This can be useful
+            when callbacks update their own state in response to their own emissions,
+            and the original emission loop is no longer relevant.
```
jni, line 127: I gotta say I don't really grok any of these, and the test is too artificial to help. Some documentation here with actual nested evented models would help here.

Anyway, as to the question of the name, I'm not sure, maybe "replace"? "immediate-drop" is certainly not intuitive as fresh-but-tired-eyes.

tlambert03: thanks for weighing in.  this has been a bit of an ongoing conversation, so the best way to understand this is definitely the test (i don't think it's too artificial if you look closer) the appending to `a` is just to make assertions easier, but lets use prints instead:

```python
from psygnal import SignalInstance

# a signal that emits an integer
sig = SignalInstance((int,), recursion_mode="deferred")


def cb1(value: int) -> None:
    print(f"calling cb1 with: {value}")
    if value == 1:
        # cb1 ALSO triggers an emission of the value 2
        sig.emit(2)

def cb2(value: int) -> None:
    print(f"calling cb2 with: {value}")
    if value == 2:
        # cb2 ALSO triggers an emission of the value 3
        sig.emit(3)

def cb3(value: int) -> None:
    print(f"calling cb3 with: {value}")


sig.connect(cb1)
sig.connect(cb2)
sig.connect(cb3)
sig.emit(1)
```

with `recursion_mode="deferred"` above: you see a breadth first pattern:
ALL callbacks are called with the first emitted value, before ANY of them are called with the second emitted value (emitted by the first connected callback  `cb1`)

```
calling cb1 with: 1
calling cb2 with: 1
calling cb3 with: 1
calling cb1 with: 2
calling cb2 with: 2
calling cb3 with: 2
calling cb1 with: 3
calling cb2 with: 3
calling cb3 with: 3
```


if you change the first line to `sig = SignalInstance((int,), recursion_mode="immediate")` you see a depth first pattern:
signal emission from within any of the callbacks *immediately* goes into the next deeper nested loop of emission events, **before returning back to the original loop to call the later callbacks with the original value**


```
calling cb1 with: 1
calling cb1 with: 2
calling cb2 with: 2
calling cb1 with: 3
calling cb2 with: 3
calling cb3 with: 3
calling cb3 with: 2
calling cb2 with: 1
calling cb3 with: 1
```

With `recursion_mode="immediate-drop"`, you're simply saying: "if, during the emission loop, a callback also emits the same signal, don't bother calling any later callbacks in this loop with that original value.  **drop** them"

```
calling cb1 with: 1
calling cb1 with: 2
calling cb2 with: 2
calling cb1 with: 3
calling cb2 with: 3
calling cb3 with: 3
```


this is useful in EventedModel because that original value (`1`) is now "stale" after cb1 emits 2, and cb2 emits 3.

tlambert03: i think the concept is straightforward enough, but the naming is indeed hard.  i called it "immediate-drop" because its the same as immediate (you immediately begin calling callbacks) but with the addition of "dropping" later callbacks in the outer loop.

please do give it a little thought, (i.e. speak now or forever hold your peace :joy:)

Czaki: Maybe this explanation should be putted in module level docstring?

tlambert03: can do.  just thinking aloud of some other names:
- `immediate-exit` (as in, immediately enter the deeper nested loop, and exit the current loop)
- `immediate-terminal` (immediately and terminally enter the nested loop)
- `terminal` (same as above: nested/recursive emissions are "terminal" ... you never come back from them)

i guess i kinda like `terminal`?

Czaki: Maybe "drop-previous-event"?

tlambert03: yep, I could go with that one too.  Will give @jni one more day to read the above

jni: > i don't think it's too artificial if you look closer

What? It's *absurdly* artificial! One would never write such a pattern. You've got (effectively) globals everywhere (`a` and `sig`) and callbacks directly calling the *same* signal with a different value. It is honestly extremely alien to me.

> you see a depth first pattern

I actually get the implications of the different modes on the traversal order. I just don't get in what situations I would ever use them or care about them. I have a vague idea that this is somehow related to nested evented model attributes in evented models (?), but those have in the past been dealt with (ineffectively, I'll admit) with blocking.

I presume that there is a real code problem that prompted you to add this feature, and that it did not involve appending to a global list — can you describe the problem that you are solving? imho that would be the best way to come up with a good name here.

jni: Thanks! That's indeed much clearer. I can see how you could end up in this situation, *and* how this mode actually makes the most sense. 🙏 

Ok now I need to sleep on it. 😂 I still find replace (you're replacing the emit loop with the new one) more intuitive than immediate-drop.

I think maybe "recursion" is the wrong way to think about it. It's kinda leaking low level details. An alternate way to think about it (not happy with exact words yet): "on_re_emission:" "enqueue" "emit_anyway" "clear".

(I'm now struggling to come up with a situation where "emit_anyway" is the right behaviour?)

tlambert03: great!  

> I still find replace (you're replacing the emit loop with the new one) more intuitive than immediate-drop.

ok, let me ponder that as well.  I think my current favorite might actually be `recursion_mode='terminal'` (i.e. if you recurse, you never come back from it).    but I can see where you're going with replace

> I think maybe "recursion" is the wrong way to think about it. It's kinda leaking low level details.

also an interesting point

tlambert03: because this is annoying me :) i put it to chat GPT.... who suggested the parameter `emission_strategy` with values:

- `sequential`: This value makes the system process all signals at the current level before moving to signals emitted by callbacks. This strategy ensures that every callback is called with every signal in the order the signals were emitted, but at their respective levels of emission depth.
- `interleaved`: This value allows callbacks to immediately emit and process new signals, intermixing their emissions with the processing of previously emitted signals. However, it retains some form of return to earlier emissions, creating a more intertwined execution order.
- `depth-first` or `no-backtrack`: Here, once a callback emits a new signal, the system focuses on processing this and any subsequently emitted signals, without returning to process any remaining signals at the previous level. This strategy suggests a depth-first execution without backtracking.

do any of things spark joy? :)

tlambert03: ok folks.  I've changed it from `recursion_model` to `emission_strategy`, and this is now the docstring.  Will wait for @jni to wake up tonight... but then merging.  cause it's starting to feel like bike shedding on a feature that no one will probably ever change :joy:

```
    emission_strategy : Literal['nested', 'sequential', 'depth-first'] | None
        Determines the order and manner in which connected callbacks are invoked when a
        callback re-emits a signal. Default is `"nested"`.

        * `"nested"`: Signals emitted by callbacks are immediately processed in
          a deeper emission loop, before returning to process signals emitted at the
          current level.

        * `"sequential"`: Ensures *all* connected callbacks are called with the first
          emitted value, before *any* of them are called with values emitted while
          calling callbacks. Signals emitted by callbacks are enqueued for emission
          after the current level of emission is complete.

        * `"depth-first"`: This is similar to `"nested"`, but control never returns to
          the previous level of emission. When a callback emits a signal, a deeper
          level of emission is started, and the current level is discontinued.
          This strategy follows a depth-first execution without backtracking.
```

jni: I like emission_strategy 🙏 (maybe reemission_strategy, but that's more awkward to type so could go either way), but I dislike the value names; particularly `depth-first`, because it implies `breadth-second` which is not the case here. 😂 I would suggest:

- nested -> immediate / in_order
- sequential -> enqueue(d) / queue(d)
- depth-first -> replace / discard / clear / most_recent_only / latest_only

Thanks for your patience. 😊

jni: (Another potential name for the parameter: `reemission='enqueued'`, `reemission='most_recent_only'` — I kinda like it.)

tlambert03: Thanks!  I will pick from those new value options and change it 👍

tlambert03: I think I like reemission too!

PR #299, opened 2024-03-12
feat: add signal aliases on SignalGroup

getzze:
Split from #265
solves https://github.com/pyapp-kit/psygnal/issues/261

Add a keyword argument to `SignalGroupDescriptor` to specify signal aliases (if the alias is None, no signal is created for this field). The signal aliases table is stored in the `SignalGroup` class (or subclass).

### DIFF
--- a/src/psygnal/_group.py
+++ b/src/psygnal/_group.py
@@ -260,6 +260,7 @@ class SignalGroup:
     _psygnal_signals: ClassVar[Mapping[str, Signal]]
     _psygnal_uniform: ClassVar[bool] = False
     _psygnal_name_conflicts: ClassVar[set[str]]
+    _psygnal_aliases: ClassVar[Mapping[str, str | None]]
 
     _psygnal_instances: dict[str, SignalInstance]
 
@@ -280,7 +281,11 @@ class SignalGroup:
         }
         self._psygnal_relay = SignalRelay(self._psygnal_instances, instance)
 
-    def __init_subclass__(cls, strict: bool = False) -> None:
+    def __init_subclass__(
+        cls,
+        strict: bool = False,
+        signal_aliases: Mapping[str, str | None] = {},
+    ) -> None:
         """Collects all Signal instances on the class under `cls._psygnal_signals`."""
         # Collect Signals and remove from class attributes
         # Use dir(cls) instead of cls.__dict__ to get attributes from super()
@@ -327,6 +332,9 @@ class SignalGroup:
                 UserWarning,
                 stacklevel=2,
             )
+
+        aliases = getattr(cls, "_psygnal_aliases", {})
+        cls._psygnal_aliases = {**aliases, **signal_aliases}
 
         cls._psygnal_uniform = _is_uniform(cls._psygnal_signals.values())
         if strict and not cls._psygnal_uniform:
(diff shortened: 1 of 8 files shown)


getzze:
All tests pass, there is just a small performance regression. This is good to review.

tlambert03:
at first glance I'm impressed!  will review more thoroughly soon.  thanks again @getzze

src/psygnal/_group.py
```diff
@@ -403,6 +411,12 @@ def __repr__(self) -> str:
         name = self.__class__.__name__
         return f"<SignalGroup {name!r} with {len(self)} signals>"
 
+    def get_signal_by_alias(self, name: str) -> SignalInstance | None:
+        sig_name = self._psygnal_aliases.get(name, name)
+        if sig_name is None or sig_name not in self:
+            return None
+        return self[sig_name]
```
tlambert03, line 418: hey @getzze, I think this is probably my only question (in an extremely nice PR).  I can see why you did it this way as opposed to modifing `__getitem__`  It allows someone to query both the alias name and the original name.  Do you think that's important? (I assume so, since you've been giving it a lot of thought and opted for this).

tlambert03:
hey @getzze, I was able to avoid the slight performance regression with https://github.com/pyapp-kit/psygnal/pull/299/commits/64e2fb4740816af427cc0408716e72f58387a556

i know it was a very small one, but time to set attribute on evented models is probably the benchmark i am most concerned about keeping fast.

src/psygnal/_group.py
getzze: hey, I think it can be removed, I mean inlined in `evented_setattr`, as it is only used there. And `SignalGroup` has one less method.

tlambert03: i do think that end-users should also be able to search by alias.  but i guess the question is should they no longer be able to search by the original name?  i.e. if I remove the method on SignalGroup, then end-users can no longer access the original names for an aliased signal, correct?  are we ok with that?

getzze:
I see you had to do something quite convoluted to avoid the performance regression. But I'm glad you managed to make it work.
I don't know how to run the benchmarks in local, so I cannot test it, but maybe it could be as fast if `with_aliases = bool(group._psygnal_aliases)` is defined inside `_setattr_and_emit_`, before line 372. So it would simplify the code for maintenance.

I can commit the change I have in mind for the benchmark workflow to run if you want.

tlambert03:
> maybe it could be as fast if with_aliases = bool(group._psygnal_aliases) is defined inside _setattr_and_emit_, before line 372. So it would simplify the code for maintenance.

I agree!  go for it

> I don't know how to run the benchmarks in local,

the codspeed benchmarks actually can't be run in local, (only the asv ones, and this wasn't a regression in those benchmarks)... so i just guessed it would help and got lucky :)


>I can commit the change I have in mind for the benchmark workflow to run if you want.

sure!

getzze:
They can always use `group._psygnal_aliases` if they want. But I'm thinking now that it's probably useful to have an easy and user-friendly way to get the signal alias for a given field.

Maybe we can have `get_signal_by_alias` be a function (that takes a `SignalGroup` as first arg) instead of a method?

getzze:
> maybe it could be as fast if `with_aliases = bool(group._psygnal_aliases)` is defined inside `_setattr_and_emit_`

Nope...

src/psygnal/_group.py
getzze: I made a function instead of a method, with a new name `get_signal_from_field`.

tlambert03:
thanks @getzze.  I'm still conflicted on how to present the fetching of signals by alias publicly.  I don't like a top-level `get_signal_from_field` function.  I removed and will just merge as is, without a public accessor.  If you need one in the future, (i.e. outside of what the evented dataclass does for you already), let me know and we'll reconsider our options

getzze:
It sounds good, thanks!

PR #307, opened 2024-05-06
perf: let EventedSet use clear() method of underlying set

DanGonite57:
Motivated by https://github.com/napari/napari/issues/6746, https://github.com/napari/napari/pull/6895

The default `clear` method implemented by `collections.abc.MutableSet` is very slow, which I guess is due to the fact that it is not backed by the C implementation of the builtin `set`. 

I have attempted to implement a custom interface for `clear`, modelled on the existing one for `discard`. However, I don't entirely understand the intentions for the various hooks, BAILs, etc, so please do let me know if anything is inconsistent with the behaviours and styles of other methods.

bench.py
```python
from psygnal.containers import EventedSet

num = 100_000

my_set = EventedSet(range(num))
my_set.events.items_changed.connect(print)
my_set.clear()
```
Before:
```cmd
> hyperfine --warmup 1 .\bench.py
Benchmark 1: .\bench.py
  Time (mean ± σ):      4.257 s ±  0.050 s    [User: 3.640 s, System: 0.047 s]
  Range (min … max):    4.192 s …  4.345 s    10 runs
```
After:
```cmd
> hyperfine --warmup 1 .\bench.py
Benchmark 1: .\bench.py
  Time (mean ± σ):     236.3 ms ±   3.4 ms    [User: 59.9 ms, System: 31.0 ms]
  Range (min … max):   231.6 ms … 243.9 ms    12 runs
```

### DIFF
--- a/benchmarks/benchmarks.py
+++ b/benchmarks/benchmarks.py
@@ -258,7 +258,7 @@ class EventedSetSuite:
     def time_update_overlap(self, n):
         self.my_set.update(range(n // 2, n + n // 2))
 
-    def time_clear(self):
+    def time_clear(self, _):
         self.my_set.clear()
 
 
--- a/benchmarks/benchmarks.py
+++ b/benchmarks/benchmarks.py
@@ -258,6 +258,9 @@ class EventedSetSuite:
     def time_update_overlap(self, n):
         self.my_set.update(range(n // 2, n + n // 2))
 
+    def time_clear(self):
+        self.my_set.clear()
+
 
 class EventedSetWithCallbackSuite(EventedSetSuite):
     def setup(self, n):
--- a/src/psygnal/containers/_evented_set.py
+++ b/src/psygnal/containers/_evented_set.py
@@ -63,6 +63,12 @@ class _BaseMutableSet(MutableSet[_T]):
             self._do_discard(_item)
             self._post_discard_hook(_item)
 
+    def clear(self) -> None:
+        _item = self._pre_clear_hook()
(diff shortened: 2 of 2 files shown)


tlambert03:
wow.  thanks this is fantastic.  LGTM

Sorry about the weird `BAIL` stuff... had to remind myself why it's there (clearly not obvious), I think it was mostly there for things like `add` where subclasses might want the ability to cancel the event, or to avoid an event entirely as you've done in the case if `len(self) == 0`.   but your PR is perfect, thanks!

tlambert03:
btw, the codspeed benchmarks shown above don't capture this, but the asv benchmarks do and are awesome 🤩 


| Change   | Before [b0c29526] <main>   | After [3b7af9d8]    |   Ratio | Benchmark (Parameter)                                     |
|----------|----------------------------|---------------------|---------|-----------------------------------------------------------|
| -        | 1.15±0.01μs                | 689±7ns             |    0.6  | benchmarks.EventedSetWithCallbackSuite.time_clear(10)     |
| -        | 1.16±0.03μs                | 682±2ns             |    0.59 | benchmarks.EventedSetSuite.time_clear(10)                 |
| -        | 4.83±0s                    | 763±2ns             |    0    | benchmarks.EventedSetSuite.time_clear(100000)             |
| -        | 4.84±0.02s                 | 794±7ns             |    0    | benchmarks.EventedSetWithCallbackSuite.time_clear(100000) |

"infinite speedup" :joy:

PR #316, opened 2024-06-19
feat: Add simple validator metadata for use in annotated

tlambert03:
Playing around with a pattern to add simple validation via `Annotated` on evented dataclasses.  There are *no* builtin validators.  It simply allows someone to declare some simple validation prior to event emission.  The use case would be when all you want is a way to cast fields without depending on a heavier library like pydantic. 

```python
    def positive_int(value: Any) -> int:
        try:
            _value = int(value)
        except (ValueError, TypeError):
            raise ValueError("Value must be an integer") from None
        if not _value > 0:
            raise ValueError("Value must be positive")
        return _value

    @evented
    @dataclass
    class Foo:
        x: Annotated[int, Validator(positive_int)]
```

(btw, @d-v-b, this is a general that might work for you in zarr.  The trick for you would just be determining when where to patch the `__setattr__` method of the parent class)

### DIFF
--- a/src/psygnal/_group_descriptor.py
+++ b/src/psygnal/_group_descriptor.py
@@ -6,8 +6,10 @@ import operator
 import sys
 import warnings
 import weakref
+from dataclasses import dataclass
 from typing import (
     TYPE_CHECKING,
+    Annotated,
     Any,
     Callable,
     ClassVar,
@@ -15,9 +17,12 @@ from typing import (
     Literal,
     Mapping,
     Optional,
+    Sequence,
     Type,
     TypeVar,
     cast,
+    get_args,
+    get_origin,
     overload,
 )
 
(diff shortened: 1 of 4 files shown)


src/psygnal/_group_descriptor.py
```diff
@@ -656,3 +705,50 @@ def _create_group(self, owner: type) -> type[SignalGroup]:
 
         self._do_patch_setattr(owner, with_aliases=bool(Group._psygnal_aliases))
         return Group
+
+
+@dataclass
+class Validator:
+    """Annotated metadatax that marks a function validates a value before setting.
```
d-v-b, line 712:
```suggestion
    """Annotated metadata marking that a function validates a value before setting.
```
not sure if this captures the intent

tlambert03:
changed my mind on this feature.  scope creepy

PR #323, opened 2024-09-06
CI: Update host OS version, add Python 3.13

EwoutH:
Some CI maintenance:
- Update the Pypi.yml host OSes to their latest compatible versions
- Add Python 3.13 to the test matrix.
  - This needs https://github.com/pyapp-kit/workflows/pull/10

### DIFF
--- a/.github/workflows/pypi.yml
+++ b/.github/workflows/pypi.yml
@@ -56,12 +56,12 @@ jobs:
         include:
           - os: ubuntu-latest
             name: linux-x86_64
-          - os: windows-2019
+          - os: windows-latest
             name: windows-amd64
-          - os: macos-11
+          - os: macos-13
             name: macos-x86_64
             macos_arch: "x86_64"
-          - os: macos-11
+          - os: macos-14
             name: macos-arm64
             macos_arch: "arm64"
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -46,6 +46,8 @@ jobs:
           # still working on ci test errors
           - os: windows-latest
             python-version: "3.12"
+          - compile: true
+            python-version: "3.13"
 
   test-qt:
     name: Test Qt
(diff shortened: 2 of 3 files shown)


EwoutH:
If preferred, feel free to work further on this branch!

tlambert03:
thanks, yeah I'm starting to look into it today.  will certainly take some CI adjustment, and it may only be possible to ship the non-mypyc (pure python) wheels at 3.13 release time (mypyc often trails python release dates slightly).  will update!

.github/workflows/pypi.yml
```diff
@@ -56,12 +56,12 @@ jobs:
...
             name: linux-x86_64
-          - os: windows-2019
+          - os: windows-latest
             name: windows-amd64
-          - os: macos-11
+          - os: macos-13
             name: macos-x86_64
             macos_arch: "x86_64"
-          - os: macos-11
+          - os: macos-14
```
tlambert03, line 64: @EwoutH, can you comment on these changes?  will these be needed for 3.13?

EwoutH: Not needed an sich, but it's good to compile on the latest host OS versions with the latest toolsets, especially for arm64.

tlambert03: huh, i was previously encouraged to compile platform specific stuff on the oldest compatible toolset...  am I misapplying/misunderstanding old advice?

EwoutH: Generally you use the latest tooling / toolset / toolchain, but with, if needed, a certain target version, which guarantees compatible with that version and up. For example, CMake uses [`MACOSX_DEPLOYMENT_TARGET`](https://cmake.org/cmake/help/latest/envvar/MACOSX_DEPLOYMENT_TARGET.html) for this.

For Python I don't believe it really matters much.

tlambert03: Yeah I remember futzing with macOS deployment target.  Can't remember if it was this project or another, but there was some issue with hatchling not always respecting that env var.  May have been fixed by now.

Anyway, let's just go with it and I'll revisit later if anything fails when it's time to release.  The main python functionality is all working fine on 3.13

Thanks again

Issue #330, opened 2024-10-23
 Platform-specific Subclassing Issue with psygnal Integration in ImSwitch


beniroquai:
* psygnal version: latest 0.11.1
* Python version: 3.10
* Operating System: Mac OS (ARM64), Raspberry Pi OS (ARM), Ubuntu 24.10 (x86); all in docker 

## Issue Description:

There seems to be a platform-specific behavior with psygnal on x86 vs ARM architectures. The problem occurs when attempting to subclass a psygnal object after integrating websockets in the ImSwitch project. The error only appears on x86 (such as on a Kubernetes Cluster) but works fine on ARM64 platforms like Raspberry Pi and MacOS, presumably because the compiled psygnal binary is installed differently between architectures.

Stack trace:

```plaintext
Starting Imswitch with the following parameters:
--headless --no-ssl --http-port 8001 --config-folder None --config-file example_virtual_microscope.json --ext-data-folder None
ERROR:root:Traceback (most recent call last):
File "/tmp/ImSwitch/imswitch/__main__.py", line 79, in main
from imswitch.imcommon import prepareApp, launchApp
File "/tmp/ImSwitch/imswitch/imcommon/__init__.py", line 1, in <module>
from .applaunch import prepareApp, launchApp
File "/tmp/ImSwitch/imswitch/imcommon/applaunch.py", line 6, in <module>
from .model import dirtools, pythontools, initLogger
File "/tmp/ImSwitch/imswitch/imcommon/model/__init__.py", line 1, in <module>
from .SharedAttributes import SharedAttributes
File "/tmp/ImSwitch/imswitch/imcommon/model/SharedAttributes.py", line 3, in <module>
from imswitch.imcommon.framework import Signal, SignalInterface
File "/tmp/ImSwitch/imswitch/imcommon/framework/__init__.py", line 4, in <module>
from .noqt import *
File "/tmp/ImSwitch/imswitch/imcommon/framework/noqt.py", line 151, in <module>
class Thread(abstract.Thread):
File "/tmp/ImSwitch/imswitch/imcommon/framework/noqt.py", line 153, in Thread
_started = Signal()
TypeError: interpreted classes cannot inherit from compiled
```

## Steps to Reproduce:
1. Run the code on an x86 platform (e.g., a Kubernetes Cluster).
2. Use the socket integration in the ImSwitch project where subclassing a psygnal object is involved.
3. Observe the above error on x86.

Links:
[...]

PR #331, opened 2024-11-04
fast: add `emit_fast` method for 10x faster emission (without safety checks)

tlambert03:
This PR is in response to @beniroquai's https://github.com/pyapp-kit/psygnal/issues/330#issuecomment-2436321438:

> but the frame-rate drops significantly in the psygnal case

Indeed, the compiled version of psygnal is roughly 5 times slower than Qt (the uncompiled version is about 6.5-7 times slower).  It looks like all that slowdown comes from safety mechanisms and other "features" that may not be used in many cases, (like the ability to query the sender, etc...).  So, this PR adds a stripped down `emit_fast()` method that has the following changes:

  - It does not support `check_nargs` or `check_types`
  - It does not use any thread safety locks.
  - It is not possible to query the emitter with `Signal.current_emitter()`
  - It is not possible to query the sender with `Signal.sender()`
  - It does not support "queued" or "latest-only" reemission modes for nested
    emits. It will always use "immediate" mode, wherein signals emitted by
    callbacks are immediately processed in a deeper emission loop.

  It DOES, however, respect `paused()` and `blocked()`

By stripping away those less-used features, the emission is about 10x faster (making the compiled version ~2x faster than Qt).  It should be used with caution, but for most simple cases, it is probably everything one needs.

### DIFF
--- a/src/psygnal/_signal.py
+++ b/src/psygnal/_signal.py
@@ -1175,6 +1175,52 @@ class SignalInstance:
 
         self._run_emit_loop(args)
 
+    def emit_fast(self, *args: Any) -> None:
+        """Fast emit without any checks.
+
+        This method can be up to 10x faster than `emit()`, but it lacks most of the
+        features and safety checks of `emit()`. Use with caution.  Specifically:
+
+        - It does not support `check_nargs` or `check_types`
+        - It does not use any thread safety locks.
+        - It is not possible to query the emitter with `Signal.current_emitter()`
+        - It is not possible to query the sender with `Signal.sender()`
+        - It does not support "queued" or "latest-only" reemission modes for nested
+          emits. It will always use "immediate" mode, wherein signals emitted by
+          callbacks are immediately processed in a deeper emission loop.
+
+        It DOES, however, support `paused()` and `blocked()`
+
+        Parameters
+        ----------
+        *args : Any
+            These arguments will be passed when calling each slot (unless the slot
+            accepts fewer arguments, in which case extra args will be discarded.)
+        """
+        if self._is_blocked:
+            return
+
+        if self._is_paused:
+            self._args_queue.append(args)
+            return
+
+        try:
+            for caller in self._slots:
+                caller.cb(args)
+        except RecursionError as e:
+            raise RecursionError(
+                f"RecursionError when "
+                f"emitting signal {self.name!r} with args {args}"
+            ) from e
+        except Exception as cb_err:
+            if isinstance(cb_err, EmitLoopError):
+                raise cb_err
+            loop_err = EmitLoopError(exc=cb_err, signal=self).with_traceback(
+                cb_err.__traceback__
+            )
+            # this comment will show up in the traceback
+            raise loop_err from cb_err  # emit() call ABOVE || callback error BELOW
+
     def __call__(
         self, *args: Any, check_nargs: bool = False, check_types: bool = False
     ) -> None:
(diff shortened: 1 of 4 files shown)


src/psygnal/_signal.py
```diff
@@ -1175,6 +1175,52 @@ def emit(
...
+                f"emitting signal {self.name!r} with args {args}"
+            ) from e
+        except Exception as cb_err:
+            if isinstance(cb_err, EmitLoopError):
+                raise cb_err
+            loop_err = EmitLoopError(exc=cb_err, signal=self).with_traceback(
+                cb_err.__traceback__
+            )
+            # this comment will show up in the traceback
+            raise loop_err from cb_err  # emit() call ABOVE || callback error BELOW
```
Czaki, line 1222:
```suggestion
        except EmitLoopError as e:
            raise e
        except Exception as cb_err:
            loop_err = EmitLoopError(exc=cb_err, signal=self).with_traceback(
                cb_err.__traceback__
            )
            # this comment will show up in the traceback
            raise loop_err from cb_err  # emit() call ABOVE || callback error BELOW
```

src/psygnal/_signal.py
```diff
@@ -1175,6 +1175,52 @@ def emit(
...
+            return
+
+        try:
+            for caller in self._slots:
+                caller.cb(args)
+        except RecursionError as e:
+            raise RecursionError(
+                f"RecursionError when "
+                f"emitting signal {self.name!r} with args {args}"
+            ) from e
```
Czaki, line 1214: I'm not a big fan of this solution, as it sometimes produces unreadable stack trace. But I understand that it is just a copy from plain `emmit` and should be fixed in both places?

Czaki (approved):
Looks good. 
did you want to add some tests to make codecov more happy? 

In not, then maybe add `#pragma: no cov` to remove noisy reports?

src/psygnal/_signal.py
tlambert03: Please suggest a better solution.  You've complained about the stack traces in the past, and I tried hard to make what I thought were useful, readable traces. I didn't know you still didn't like them, but it would be good to have some more concrete suggestions

beniroquai:
This sounds great! Thanks a lot for all the effort! Is it an in-place upgrade or would I need to tweak any of the settings? 
CC: @jacopoabramo

Czaki:
You need to call `emit_fast` instead of `emit` so it may require some subclassing if you want to use the same interface as in QSignal.

jacopoabramo:
> This sounds great! Thanks a lot for all the effort! Is it an in-place upgrade or would I need to tweak any of the settings? CC: @jacopoabramo

I guess you can just replace the call to `emit` with `fast_emit` within the child class. The base API doesn't change.

jacopoabramo:
Out of curiosity, how does `emit_fast` drop support for thread safety locks? From the PR it doesn't that there's anything related to threading management but I don't know the whole code base very well

tlambert03:
> Is there a proper way to enforce --no-binary when installing psygnal through the setup.py

@beniroquai ... no there's no way to enforce that in your dependencies, since that is a pip-specific option at install time, and not related to project metadata.  You can, however, tell pip to install with no_binary for a specific package at install time:

```sh
PIP_NO_BINARY=psygnal pip install ImSwitch 
```

Alternatively, you can do the "decompilation" (which simply renames the compiled libraries) in your source code, next to where you subclass, though this requires write access to the psygnal files in site-packages:

```python
# somewhere in your ImSwitch code
import psygnal.utils
psygnal.utils.decompile()
```

... I'd call that a short-term patch, until I have time to look closer at #330

PR #335, opened 2024-11-08
Update EventedModel docs and errors to refer `field_dependencies`

sjdemartini:
`property_dependencies` was renamed to `field_dependencies` in https://github.com/pyapp-kit/psygnal/pull/234 but some associated error messages and docs still refer to the deprecated name.


tlambert03 (approved):
oops, thank you!

PR #336, opened 2024-11-08
fix: fix use of computed_field setter with field_dependencies

### LINKED ISSUE #334: Adding Pydantic v2 `@computed_field` decorator prevents class creation with dependencies
* psygnal version: 0.11.1
* Python version: 3.10
* Operating System: MacOS

### Description

(First of all, thank you for building this excellent library! It's been super useful and a much nicer alternative to param, traitlets, etc!)

If we add the `@computed_field` decorator (https://docs.pydantic.dev/latest/concepts/fields/#the-computed_field-decorator) to any property which we want to be an evented Psygnal `@property` field that is part of `field_dependencies`, it raises an error and prevents definition of the class. 

`@computed_field` is useful, since it ensures the field appears in Pydantic's model serialization, and I would expect it to work with `EventedModel`s as well (esp since Psygnal docs refer to supporting "computed fields" for this sort of use-case).

**Side note/ask**: It would be nice if the `@computed_field` could be added to the `model.events` `SignalGroup` even _without_ a setter, so that at least we could manually `model.events.my_computed_field.emit()` too.

### What I Did

Following the example similar to the docs (https://psygnal.readthedocs.io/en/stable/API/model/), adding `@comptued_field` to the `c` property, as follows:

```python
from pydantic import computed_field

class MyModel(EventedModel):
    a: int = 1
    b: int = 1

    @computed_field
    @property
    def c(self) -> List[int]:
        return [self.a, self.b]

    @c.setter
    def c(self, val: Sequence[int]) -> None:
        self.a, self.b = val

    class Config:
        allow_property_setters = True
        field_dependencies = {"c": ["a", "b"]}
```

This leads to an error:

```
----> 3 class MyModel(EventedModel):
      4     a: int = 1
      5     b: int = 1

File /path/to/lib/python3.10/site-packages/psygnal/_evented_model.py:280, in EventedMetaclass.__new__(mcs, name, bases, namespace, **kwargs)
    274             if conf and conf.get(ALLOW_PROPERTY_SETTERS, False):
    275                 raise ValueError(
    276                     "Cannot set 'allow_property_setters' to 'False' when base "
    277                     f"class {b} sets it to True"
    278                 )
--> 280 cls.__field_dependents__ = _get_field_dependents(
    281     cls, model_config, model_fields
    282 )
    283 cls.__signal_group__ = type(f"{name}SignalGroup", (SignalGroup,), signals)
    284 if not cls.__field_dependents__ and hasattr(cls, "_setattr_no_dependants"):

File /path/to/lib/python3.10/site-packages/psygnal/_evented_model.py:337, in _get_field_dependents(cls, model_config, model_fields)
    335 for prop, fields in cfg_deps.items():
    336     if prop not in {*model_fields, *cls.__property_setters__}:
--> 337         raise ValueError(
    338             "Fields with dependencies must be fields or property.setters."
    339             f"{prop!r} is not."
    340         )
    341     for field in fields:
    342         if field not in model_fields:

ValueError: Fields with dependencies must be fields or property.setters.'c' is not.
```

tlambert03:
fixes #334

### DIFF
--- a/src/psygnal/_evented_model.py
+++ b/src/psygnal/_evented_model.py
@@ -12,7 +12,6 @@ from typing import (
     NamedTuple,
     Set,
     Type,
-    TypeGuard,
     Union,
     cast,
     no_type_check,
@@ -34,6 +33,7 @@ if TYPE_CHECKING:
     from pydantic._internal import _model_construction as pydantic_main
     from pydantic._internal import _utils as utils
     from pydantic._internal._decorators import PydanticDescriptorProxy
+    from typing_extensions import TypeGuard  # py310
     from typing_extensions import dataclass_transform as dataclass_transform  # py311
 
     from ._signal import SignalInstance
@@ -135,7 +135,7 @@ if not PYDANTIC_V1:
     def _model_dump(obj: pydantic.BaseModel) -> dict:
         return obj.model_dump()
 
-    def _is_pydantic_descriptor_proxy(obj: Any) -> TypeGuard["PydanticDescriptorProxy"]:
+    def _is_pydantic_descriptor_proxy(obj: Any) -> "TypeGuard[PydanticDescriptorProxy]":
         if (
             type(obj).__module__.startswith("pydantic")
             and type(obj).__name__ == "PydanticDescriptorProxy"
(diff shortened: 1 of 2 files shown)


sjdemartini:
Awesome, thank you for adding this! Looks great from the tests!

PR #340, opened 2024-11-21
feat: support pydantic @computed_fields,  and arbitrary `field_dependents` for event emission

### LINKED ISSUE #337: Adding additional signals (e.g. for computed_field)
from @sjdemartini in #334:

> Side note/ask: It would be nice if the @computed_field could be added to the model.events SignalGroup even without a setter, so that at least we could manually model.events.my_computed_field.emit() too.

see also, discussion starting here: https://github.com/pyapp-kit/psygnal/issues/334#issuecomment-2465677934

tlambert03:
This is a proof of principle PR that would support the feature that @sjdemartini is requesting in https://github.com/pyapp-kit/psygnal/issues/337.  namely:

- computed fields can gain events
- computed fields (really, currently any field) can list *non*-field attribute names as their dependents for checking event emission.

@Czaki, I would appreciate your eyes on this. 
1. I just "made it work" ... so there's likely some minor refactoring that should be done here to polish it (particularly in the `_check_if_values_changed_and_emit_if_needed` area that you had previously worked on)
2. Feel free to comment in general on the feature request.  I think @sjdemartini motivated it quite nicely his comment, https://github.com/pyapp-kit/psygnal/issues/337#issuecomment-2472157973 ... so please read that for context; and for a quick use case example, see the new test I added in `test_private_field_dependents`

closes #337


tlambert03:
tests are failing during building of mypyc, due to https://github.com/pydantic/pydantic/issues/10907

src/psygnal/_evented_model.py
```diff
@@ -565,15 +577,23 @@ def _check_if_values_changed_and_emit_if_needed(self) -> None:
             # do not run whole machinery if there is no need
             return
         to_emit = []
+        must_continue = False
         for name in self._primary_changes:
             # primary changes should contains only fields
             # that are changed directly by assignment
```
Czaki, line 583: I do not fully understand. When we need to emit a signal. This comment still statement that `self._primary_changes` are only fields changed by direct assignment. 
And documentation of https://docs.pydantic.dev/2.0/usage/computed_fields/ do not mention the assigment of this property.

Czaki (commented):
Why can we not use the same machinery like current property setter/getter (put things in `self._changes_queue`)?

src/psygnal/_evented_model.py
tlambert03: yeah, i wasn't entirely clear on the `_primary_changes`, and `_changes_queue` attributes here.  So maybe I did it wrong.  we need to emit a signal here when a *non-field* attribute changes, provided that some other evented field lists that attribute as one of its "dependents".  So, in the example provided by the original requester:

```python
from pydantic import computed_field, PrivateAttr

class MyModel(EventedModel):
    _items_dict: dict[str, int] = PrivateAttr(default_factory=dict)

    @computed_field
    def item_names(self) -> list[int]:
        return list(self._items_dict.keys())

    def add_item(self, name: str, value: int) -> None:
        # this next assignment should trigger the emission of `self.events.item_names`
        # because `_items_dict` was listed as a field_dependency of `item_names`
        self._items_dict = {**self._items_dict, name: value}

    class Config:
        field_dependencies = {"item_names": ["_items_dict"]}
```

if I did that incorrectly (or stupidly) with the existing `_primary_changes/_changes_queue` pattern let me know

tlambert03: i think this might be the first case where the "primary change" is to a non-evented attribute (`_items_dict`)... and so it didn't quite fit into the existing pattern.  It's the first case where a `_primary_change` isn't actually able to itself be evented

Czaki: Hm. 

Maybe we should add a dummy emiter (callback that do nothing, and raises an exception on connection) for private attributes that are listed in `field_dependencies`? Then we will move the complexity from setattr to constructor call?

tlambert03: that's worth trying, lemme see how it goes

PR #346, opened 2024-12-20
feat: support coroutine functions

### LINKED ISSUE #345: Connect signal to async callback
* psygnal version: 0.11.1
* Python version: 3.13.1
* Operating System: Debian 12

### Description

I'd like to connect a signal to an async callback, but it doesn't seem to be supported.

### What I Did

This gives a `"coroutine 'on_value_changed' was never awaited"` warning:
```py
from psygnal import Signal

# define an object with class attribute Signals
class MyObj:

    # this signal will emit a single string
    value_changed = Signal(str)

    def __init__(self, value=0):
        self._value = value

    def set_value(self, value):
        if value != self._value:
            self._value = str(value)
            # emit the signal
            self.value_changed.emit(self._value)

async def on_value_changed(new_value: str):
    print(f"The new value is {new_value!r}")

# instantiate the object with Signals
obj = MyObj()

# connect one or more callbacks with `connect`
obj.value_changed.connect(on_value_changed)

# callbacks are called when value changes
obj.set_value('hello!')  # prints: 'The new value is 'hello!'
```

tlambert03:
fixes #345 

See https://psygnal.readthedocs.io/en/latest/guides/async/ for details on usage


davidbrochart:
Thanks for the quick PR @tlambert03!
In #345 you mentioned that:
> psygnal itself has no concept of an event loop (it's a "bring-your-own-event-loop" library)

But this PR currently assumes that the event loop is asyncio. However Trio and AnyIO are very popular, so I'm wondering how we could support them and really be event-loop agnostic.

davidbrochart:
Because structured concurrency always needs a task group in order to launch tasks, I think it's better to not support async callbacks in psygnal. One can just use a sync callback, and have the async logic in the user code where the new value is put synchronously in a stream, and a background task gets the values from the stream asynchronously. That way the user can launch this background task in the appropriate task group.

tlambert03:
Thanks for your input!

> I think it's better to not support async callbacks in psygnal. One can just use a sync callback, and have the async logic in the user code where the new value is put synchronously in a stream, and a background task gets the values from the stream asynchronously.

Are you sure?  I'm definitely happy to do whatever we can here to make using coroutines in psygnal easier.  That could include supporting auto detection of asyncio, trio, or anyio.

I don't have a ton of personal experience with async/await, so would happily take advice and implement whatever folks suggest.
At the very least we could add some better documentation explaining/showing the strategy you mentioned for using a sync function to put a task in a stream.

And if there are conveniences we could add around that pattern, that's fine too

tests/test_coroutines.py
```diff
@@ -0,0 +1,80 @@
...
+        "weak_coroutinefunc",
+        "coroutinemethod",
+    ],
+)
+@pytest.mark.asyncio
+async def test_slot_types(type_: str, capsys: Any) -> None:
+    backend = _async.set_async_backend("asyncio")
+    assert backend is _async.get_async_backend() is not None
+
+    await asyncio.sleep(0)
```
davidbrochart, line 25:
```suggestion
    while True:
        if backend.running:
            break
        await asyncio.sleep(0)
```

tlambert03: thanks.... yes, I'm going to have a number of best-practice questions for you :) i really don't use this stuff much.  just a moment and I will post more questions

src/psygnal/_weak_callback.py
```diff
@@ -580,3 +596,37 @@ def cb(self, args: tuple[Any, ...] = ()) -> None:
...
+
+
+class WeakCoroutineFunction(WeakFunction):
+    def cb(self, args: tuple[Any, ...] = ()) -> Any:
+        if self._f() is None:
+            raise ReferenceError("weakly-referenced object no longer exists")
+        if self._max_args is not None:
+            args = args[: self._max_args]
+
+        cast("_AsyncBackend", get_async_backend())._put((self, args))
```
tlambert03, line 611: @davidbrochart one change I had to make here was to put the WeakCallback itself in the queue, rather than the dereferenced object, in order to avoid holding strong references.  (minor detail... but something this library tries hard to avoid for reasons of having been bitten by it in the past).  See corresponding change in `_AsyncBackend.call_back`

tests/test_coroutines.py
```diff
@@ -0,0 +1,80 @@
...
+    "type_",
+    [
+        "coroutinefunc",
+        "weak_coroutinefunc",
+        "coroutinemethod",
+    ],
+)
+@pytest.mark.asyncio
+async def test_slot_types(type_: str, capsys: Any) -> None:
+    backend = _async.set_async_backend("asyncio")
```
tlambert03, line 22: @davidbrochart I think one of the biggest questions for me is "proper usage".  This is what I did to get the tests to work (so far, just for asyncio).  But note that I had to allow `_async.set_async_backend("asyncio")` to be called *multiple* times within a session... (with the restriction that it be called with the same backend).

An important thing to note there though is that, in `set_async_backend`, I actually had to let it re-initialize the backend (can't just return the already instantiated backend)... otherwise I get 

```
    def _check_closed(self):
        if self._closed:
>           raise RuntimeError('Event loop is closed')
E           RuntimeError: Event loop is closed
```

on follow-up tests.


Another question I have here is how to test all three backends.  Is it possible within a single python session?

tests/test_coroutines.py
```diff
@@ -0,0 +1,80 @@
...
+
+        obj = MyObj()
+        cb = weak_callback(obj.coroutine_method, finalize=final_mock)
+
+    assert isinstance(cb, WeakCallback)
+    assert isinstance(cb.slot_repr(), str)
+    assert cb.dereference() is not None
+
+    cb.cb((2,))
+    await asyncio.sleep(0.01)
```
tlambert03, line 55: similar to your comment above... i suspect this is flaky.  Should we have an API to wait until the backend has no items in the queue?

tlambert03:
one additional note, I'm skipping the `_async.py` module from the mypyc compilation targets for now.  When I include it, it's giving me odd, seemingly-unrelated errors in other modules.  it *might* have to do with trio/anyio missing from the environment during mypyc compilation.  But I think it's probably just asking for problems anyway.   The key bits (signal emissions and weak-callbacks) are still compiled

src/psygnal/_weak_callback.py
davidbrochart: I think this is fine :+1:

tests/test_coroutines.py
davidbrochart: Maybe we should still prevent `set_async_backend()` from being called multiple times, unless a `clear_async_backend()` function is called in between (mostly for tests)?

tests/test_coroutines.py
davidbrochart: That's a possibility, but a simpler approach could be to set an event in the mock, and wait for it (with a timeout)?

src/psygnal/_async.py
```diff
@@ -0,0 +1,226 @@
...
+    def close(self) -> None:
+        """Close the asyncio backend and cancel tasks."""
+        if hasattr(self, "_task") and not self._task.done():
+            self._task.cancel()
+
+    async def run(self) -> None:
+        if self.running:
+            return
+
+        self._running = True
```
davidbrochart, line 126:
```suggestion
        self._running.set()
```

src/psygnal/_async.py
```diff
@@ -0,0 +1,226 @@
...
+                    import traceback
+
+                    traceback.print_exc()
+        except self._asyncio.CancelledError:
+            pass
+        except RuntimeError as e:
+            if not self._loop.is_closed():
+                raise e
+        finally:
+            self._running = False
```
davidbrochart, line 144:
```suggestion
            self._running.clear()
```

src/psygnal/_async.py
```diff
@@ -0,0 +1,226 @@
...
+    _send_stream: anyio.streams.memory.MemoryObjectSendStream[QueueItem]
+    _receive_stream: anyio.streams.memory.MemoryObjectReceiveStream[QueueItem]
+
+    def __init__(self) -> None:
+        super().__init__("anyio")
+        import anyio
+
+        self._send_stream, self._receive_stream = anyio.create_memory_object_stream(
+            max_buffer_size=inf
+        )
```
davidbrochart, line 157:
```suggestion
        )
        self._running = anyio.Event()
```

src/psygnal/_async.py
```diff
@@ -0,0 +1,226 @@
...
+
+    def close(self) -> None:
+        """Close the anyio streams."""
+        if hasattr(self, "_send_stream"):
+            self._send_stream.close()
+        if hasattr(self, "_receive_stream"):
+            self._receive_stream.close()
+
+    async def run(self) -> None:
+        if self.running:
```
davidbrochart, line 173:
```suggestion
        if self.running.is_set():
```

src/psygnal/_async.py
```diff
@@ -0,0 +1,226 @@
...
+            return  # pragma: no cover
+
+        self._running = True
```
davidbrochart, line 176:
```suggestion
        self._running.set()
```

src/psygnal/_async.py
```diff
@@ -0,0 +1,226 @@
...
+                async for item in self._receive_stream:
+                    try:
+                        await self.call_back(item)
+                    except Exception:
+                        # Log the exception but continue running
+                        import traceback
+
+                        traceback.print_exc()
+        finally:
+            self._running = False
```
davidbrochart, line 188:
```suggestion
            self._running = anyio.Event()
```

src/psygnal/_async.py
```diff
@@ -0,0 +1,226 @@
...
+    _send_channel: trio._channel.MemorySendChannel[QueueItem]
+    _receive_channel: trio.abc.ReceiveChannel[QueueItem]
+
+    def __init__(self) -> None:
+        super().__init__("trio")
+        import trio
+
+        self._send_channel, self._receive_channel = trio.open_memory_channel(
+            max_buffer_size=inf
+        )
```
davidbrochart, line 203:
```suggestion
        )
        self._running = trio.Event()
```

src/psygnal/_async.py
```diff
@@ -0,0 +1,226 @@
...
+
+    def _put(self, item: tuple) -> None:
+        self._send_channel.send_nowait(item)
+
+    async def _get(self) -> tuple:
+        return await self._receive_channel.receive()
+
+    async def run(self) -> None:
+        if self.running:
```
davidbrochart, line 212:
```suggestion
        if self.running.is_set():
```

src/psygnal/_async.py
```diff
@@ -0,0 +1,226 @@
...
+            return  # pragma: no cover
+
+        self._running = True
```
davidbrochart, line 215:
```suggestion
        self._running.set()
```

src/psygnal/_async.py
```diff
@@ -0,0 +1,226 @@
...
+            async for item in self._receive_channel:
+                try:
+                    await self.call_back(item)
+                except Exception:
+                    # Log the exception but continue running
+                    import traceback
+
+                    traceback.print_exc()
+        finally:
+            self._running = False
```
davidbrochart, line 226:
```suggestion
            self._running = trio.Event()
```

src/psygnal/_weak_callback.py
```diff
@@ -124,14 +129,46 @@ def _on_delete(weak_cb):
...
         cb = cb.func
 
+    is_coro = inspect.iscoroutinefunction(cb)
+    if is_coro:
+        if (backend := get_async_backend()) is None:
+            raise RuntimeError(
+                "Cannot create async callback yet... No async backend set. "
+                "Please call `psygnal.set_async_backend()` before connecting."
+            )
+        if not backend.running:
```
davidbrochart, line 139:
```suggestion
        if not backend.running.is_set():
```

tests/test_coroutines.py
```diff
@@ -0,0 +1,491 @@
...
+                _async.clear_async_backend()
+
+        return asyncio.run(_run_test())
+
+    async def _wait_for_backend_running(
+        self, backend: _async._AsyncBackend, timeout: float = 1.0
+    ) -> None:
+        """Wait for backend to be running with a timeout."""
+        start_time = asyncio.get_event_loop().time()
+        while not backend.running:
```
davidbrochart, line 86:
```suggestion
        while not backend.running.is_set():
```

tests/test_coroutines.py
```diff
@@ -0,0 +1,491 @@
...
+
+@pytest.mark.usefixtures("clean_async_backend")
+@pytest.mark.asyncio
+async def test_run_method_early_return() -> None:
+    """Test that run() method returns early if backend is already running."""
+    backend = _async.set_async_backend("asyncio")
+
+    # Wait for backend to be running
+    start_time = asyncio.get_event_loop().time()
+    while not backend.running:
```
davidbrochart, line 346:
```suggestion
    while not backend.running.is_set():
```

tests/test_coroutines.py
```diff
@@ -0,0 +1,491 @@
...
+        if asyncio.get_event_loop().time() - start_time > 1.0:
+            raise TimeoutError("Backend did not start running within timeout")
+        await asyncio.sleep(0)
+
+    # Now calling run() again should return early
+    await backend.run()
+
+    # Backend should still be running
+    assert backend.running
```
davidbrochart, line 355:
```suggestion
    assert backend.running.is_set()
```

tests/test_coroutines.py
```diff
@@ -0,0 +1,491 @@
...
+
+                async with trio.open_nursery() as nursery:
+                    # Start the backend in the nursery
+                    async_backend = _async.get_async_backend()
+                    assert async_backend is not None
+                    nursery.start_soon(async_backend.run)
+
+                    # Wait for backend to be running
+                    while not backend.running:
+                        await trio.sleep(0)
```
davidbrochart, line 413:
```suggestion
                    await backend.running.wait()
```

tests/test_coroutines.py
```diff
@@ -0,0 +1,491 @@
...
+
+                async with anyio.create_task_group() as tg:
+                    # Start the backend in the task group
+                    async_backend = _async.get_async_backend()
+                    assert async_backend is not None
+                    tg.start_soon(async_backend.run)
+
+                    # Wait for backend to be running
+                    while not backend.running:
+                        await anyio.sleep(0)
```
davidbrochart, line 393:
```suggestion
                    await backend.running.wait()
```

tests/test_coroutines.py
```diff
@@ -0,0 +1,491 @@
...
+                import asyncio
+
+                # Start the backend as a background task
+                async_backend = _async.get_async_backend()
+                assert async_backend is not None
+                task = asyncio.create_task(async_backend.run())
+
+                # Wait for backend to be running
+                while not backend.running:
+                    await asyncio.sleep(0)
```
davidbrochart, line 380:
```suggestion
                await backend.running.wait()
```

tests/test_coroutines.py
```diff
@@ -0,0 +1,491 @@
...
+            backend = _async.set_async_backend("trio")
+
+            # Use a timeout to prevent hanging
+            with trio.move_on_after(5.0) as cancel_scope:
+                async with trio.open_nursery() as nursery:
+                    nursery.start_soon(backend.run)
+
+                    # Wait for backend to be running
+                    while not backend.running:
+                        await trio.sleep(0)
```
davidbrochart, line 172:
```suggestion
                    await backend.running.wait()
```

tests/test_coroutines.py
```diff
@@ -0,0 +1,491 @@
...
+            _async.clear_async_backend()
+            backend = _async.set_async_backend("anyio")
+
+            result = None
+            async with anyio.create_task_group() as tg:
+                tg.start_soon(backend.run)
+
+                # Wait for backend to be running
+                while not backend.running:
+                    await anyio.sleep(0)
```
davidbrochart, line 121:
```suggestion
                await backend.running.wait()
```

davidbrochart (changes requested):
Thanks @tlambert03, that looks good.
I'd just like the `running` property of a backend to be an event, so that it can just be awaited.

PR #349, opened 2025-01-22
docs: modified quickstart simple example

roynielsen17:
Adding a while loop to the index.md example with the ability to show what happens with looping, seeing multiple signals sent and handled, seeing the changed print statements when the signals are handled, when the inputs are changed and signals emitted.

example tested with python 3.9+ python on macos

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -29,11 +29,19 @@ my_obj = MyObject()
 # You (or others) can connect callbacks to your signals
 @my_obj.value_changed.connect
 def on_change(new_value: str):
-    print(f"The value changed to {new_value}!")
+    print(f'Hi {new_value}!') # prints the new string - Hi {new_value}
 
-# The object may now emit signals when appropriate,
-# (for example in a setter method)
-my_obj.value_changed.emit('hi')  # prints "The value changed to hi!"
+# Mock event loop to see the effect of new signals with 
+# new values - watch the signal handler handle the changing values
+count = 0
+my_input = ""
+while count < 5:
+    my_input = input("enter a name here: ")
+
+    # The object may now emit signals when appropriate,
+    # (for example in a setter method)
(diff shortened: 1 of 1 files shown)


roynielsen17:
This submission is deliberately simpler than pyside6/Qt - to give people a very simple example of signals and slots so they don't have to learn the whole PySide/Qt Frameworks before understanding Signals and Slots. (I've programmed with them on and off for over a decade)

Here's  a more complex version that has a pytest component:

"""
Two ways to run this file:

% python changing_psygnal.py

% pytest changing_psygnal.py

"""

from psygnal import Signal
import pytest

class MyObject:
    # define one or signals as class attributes
    value_changed = Signal(str)

# create an instance
my_obj = MyObject()

# You (or others) can connect callbacks to your signals
@my_obj.value_changed.connect
def on_change(new_value: str):
    print(f'Hi {new_value}!') # prints the new string - Hi {new_value}
    my_obj.new_value = new_value

def get_name_and_say_hello():
    my_input = input("enter a name here: ")
    my_obj.value_changed.emit(str(my_input))
    return my_obj.new_value

if __name__ == "__main__":
    count = 0
    while count < 5:
        name = get_name_and_say_hello()
        print(f"Said Hi to.... {name}")
        count = count + 1

""" TEST """
def test_say_hello(monkeypatch):
    names = ['Sally', 'Bob', 'Harriet', 'Bart', 'Erica']
    count = 0
    while count < 5:
        monkeypatch.setattr('builtins.input', lambda _: names[count])
        result = get_name_and_say_hello()
        assert result == names[count]
        count = count + 1

tlambert03:
thanks for the suggestion @roynielsen17.  If you think this clarifies something about the usage pattern, I'm happy to include it

roynielsen17:
Absolutely!

Respectfully,
-Roy Nielsen

PR #354, opened 2025-02-02
refactor: have EmitLoopError inherit RuntimeError

tlambert03:
@Czaki how do you feel about this change.

I think it more accurately captures the semantics of an error during a callback, and it also is useful for try/except when you're also dealing with C++ object deleted from the qt side


tlambert03:
ah, looks like this is not easily done with mypyc

PR #379, opened 2025-07-01
feat: allow SignalInstances for evented dataclass fields to emit on field mutation (in addition to field change)

### LINKED ISSUE #375: Bubbled up events could/should trigger parent event
See discussion starting here:

https://github.com/pyapp-kit/psygnal/pull/298#issuecomment-3016241595

and converging here:

https://github.com/pyapp-kit/psygnal/pull/298#issuecomment-3016266083

Following the example in the docs: it would be good if we set up an evented dataclass as follows:

    from dataclasses import dataclass, field
    from psygnal import evented

    @evented(connect_child_events=True)  # default is True
    @dataclass 
    class Person:
        name: str = ""
        age: int = 0

    @evented(connect_child_events=True)  # default is True
    @dataclass
    class Team:
        name: str = ""
        leader: Person = field(default_factory=Person)
        
Then it would be good if when we set up:

```python
team = Team(
    name='A-team'
    leader=Person(name='Hannibal', age=59)
)
```

Then when Hannibal's birthday comes along, setting:

```python
team.leader.age = 60
```

Would trigger the callbacks connected to `team.events.leader` with the full value of `Person(name='Hannibal', age=60)`.

Currently, in order to receive bubbled up events from `leader`, one needs to subscribe to the `team.events()` group and filter out the `name` events. Additionally, `team.events.leader()` is an event that many users might reach for to be notified about changes in leader, and might expect to be triggered when an attribute in `leader` changes.

tlambert03:
closes #375

```python
@evented
@dataclass
class Person:
    name: str = ""
    age: int = 0

@evented
@dataclass
class Team:
    name: str = ""
    leader: Person = field(default_factory=Person)

team = Team(name="A-Team", leader=Person(name="Hannibal", age=59))

mock = Mock()
team.events.leader.connect(mock, emit_on_evented_child_events=True)
team.leader.age = 60
mock.assert_called_once_with(Person(name="Hannibal", age=60), None)
```

just want to express a slight degree of concern over the magic that is happening here :joy: ... i recognize that having `team.events.leader` not emit when you set `team.leader.age` *is* what @jni referred to elsewhere as an "attractive nuisance" ... but the semantics and mechanics have always been _consistent_ at least (even if it can be confusing to keep it all straight in your head what everything does): `team.events.leader` has always been emitted when `team.leader` is SET to something new (not when `team.leader` is mutated).

With this, we do absolutely begin to go down a more magical path of "I know what you probably wanted, let me just do that for you...".  I can't say for certain that this won't have unexpected consequences. 

-----

also, it's currently hard for us to retrieve the old value before the change, so while `team.events.leader` *usually* gets called back with `(new_leader, old_leader)` ... it's more difficult here for us to reconstruct the value of old_leader before the mutation occurred.  It's not that it's impossible... but it would either require a potentially error reconstruction of an old value, or temporary caching of the old_value somewhere.

so, that's why in the current example above, the callback received `None` for the second value (unlike the usual dataclass value-changed signal(

### DIFF
--- a/src/psygnal/_group.py
+++ b/src/psygnal/_group.py
@@ -346,8 +346,11 @@ class SignalRelay(SignalInstance):
                     if deref == slot:
                         return i
                     # Case 2: compiled path where we stored __call__ bound method
-                    owner = getattr(deref, "__self__", None)
-                    if isinstance(owner, _relay_partial) and owner == slot:
+                    # (these will never hit on codecov, but they are covered in tests)
+                    owner = getattr(deref, "__self__", None)  # pragma: no cover
+                    if (
+                        isinstance(owner, _relay_partial) and owner == slot
+                    ):  # pragma: no cover
                         return i
         return -1  # pragma: no cover
 


jni:
Just dug this one back up from my notifications. Any updated thoughts on this @tlambert03? I would very much like to have it. 😊 I've looked through the code and it's very nice and well documented. So only thing missing seems to be passing the tests (I haven't checked them), especially since default is False.

jni:
> I'd like to reserve the right to revert if unexpected downsides are revealed in the future :)

Fine by me! I'll play with it in "toy" projects like skan (😂) before going full bore with it in like napari. 😂 Having said that, you should probably add a note in the docstring that it could be removed! Not everyone is going to discover this feature via this PR to see your reservations!

PR #387, opened 2025-09-24
feat: Allow to replace selection content in with one event emmited

Czaki:
Add a method to replace selection content with a single `items_changed` event without needing to reimplement the `_reduce_events` function.

Fix `select_only` to also emit only single `items_changed` event


tlambert03:
all set

PR #407, opened 2026-01-02
fix: Fix leaking objects by `connect_child_events=True` in `SignalGroupDescriptor` 

Czaki:
closes #406


tlambert03:
I can release today

nan-guo:
Thanks! @tlambert03

Issue #419, opened 2026-06-03
Slow extend() for EventedList and friends


ndxmrb:
Hi,
I want to insert a batch of objects into a `SelectableEventedList`, but I want to fire corresponding signals only once, not for each object. The purpose is to reduce update calls in widgets down the line.

I thought I'd use `paused()` with a reducer to collect all indexes. But `inserting` currently only takes an `index: int`, not a tuple.

**Question**: How can I defer signal emission properly, such that Qt models update properly? 

I tried blocking, but then you need to decide what to emit when done.

Or should I rather try to throttle the model functions? 

**Suggestion**: `inserting` etc. could accept tuples as well, such that Qt's `beginInsertRows()` and `endInsertRows()` can be used to reduce update calls. `extend()` could then make use of this directly, such that it only emits the `inserted` signal once.

Thanks in advance,
ndxmrb

tlambert03:
would this work for your case?

```python
from __future__ import annotations

from collections.abc import Sequence
from qtpy.QtCore import QAbstractItemModel, QModelIndex
from psygnal.containers import EventedList


def batch_extend(model: QAbstractItemModel, lst: EventedList, items: Sequence) -> None:
    start = len(lst)
    model.beginInsertRows(QModelIndex(), start, start + len(items) - 1)
    try:
        with lst.events.inserting.blocked(), lst.events.inserted.blocked():
            lst.extend(items)
    finally:
        model.endInsertRows()
```

ndxmrb:
Hey,
thank you for the quick answer.

Yes, in principle that would work for one specific case. But it needed to be done for every model in the application and isn't a solution for widgets relying on the signals emitted from the EventedList directly.

Should such widgets throttle their update function? If so, how can the potential thread mismatch be handled, because UI elements can only be modified in the main thread?

tlambert03:
can we use some more concrete MRE examples?   I think this mostly boils down to an application question, which I'm happy to weigh in on, but I'm not sure there's a global answer here.  Some of it relates to idiosyncrasies of Qt specifically (which again, we do want to be able to "have an answer for", but backing a specific QAbstractModel isn't the sole mission of psygnal)

> it needed to be done for every model in the application and isn't a solution for widgets relying on the signals emitted from the EventedList directly.

[...]

PR #422, opened 2026-06-10
feat: Add batch signal support for contiguous item insertions and removals in EventedList

tlambert03:
playing with explicit support of batch signals, re #419

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -98,9 +98,9 @@ jobs:
     uses: pyapp-kit/workflows/.github/workflows/test-dependents.yml@main
     with:
       dependency-repo: pyapp-kit/magicgui
-      dependency-group: "pyside2"
-      qt: "pyside2"
-      python-version: "3.10"
+      dependency-group: "pyside6"
+      qt: "pyside6"
+      python-version: "3.12"
 
   typing:
     runs-on: ubuntu-latest
--- a/README.md
+++ b/README.md
@@ -105,6 +105,14 @@ my_list.events.removed.connect(lambda i, val: print(f"Removed {val} at index {i}
 
 my_list.append(6)  # Output: Inserted 6 at index 5
 my_list.pop()  # Output: Removed 6 at index 5
+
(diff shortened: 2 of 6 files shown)


ndxmrb:
Good stuff! This makes updates for big amounts of data so much easier. Thank you!

ndxmrb:
Is there anything in particular you want to add before you merge, or is it ready to go? 
The coverage report says the early return is uncovered, but other than that?

To me, it's been a great addition already and using it improves performance significantly. Would love for this to be merged and released 🙂 

From the pending `EventedModel` PR I know your quality standards are very high and there's many things to evaluate beneath the surface. So no pressure, I'm just curious!

tlambert03:
thanks for the ping @ndxmrb (and sorry again for the delay).  I think it's fine.  the only thing I'd like opinions on (from you, maybe @Czaki) is whether we like the names of the new signals `items_inserting / items_inserted` ... or whether we want something like `batch_inserting / batch_inserted`?  anyone have any opinions?  Will merge in a couple days if not

ndxmrb:
Now that you say it, I like `batch_inserting / batch_inserted` better. To me, that distinguishes the signal for the item (fired every time) from the one for the batch (fired for every contiguous batch) a bit better.

tlambert03:
yeah i think i do too.  will change and merge

ndxmrb:
Interesting bit about slices. I guess I need to check my implementation, because I only listen to the `batch_` signals currently.

ndxmrb:
Hi again,
Not to be annoying, but could this be merged and put into a new release please?

I feel like you already did all the hard work and it'd be "wasteful" (for lack of a better term) to let it get stale.

Then again, maybe you have your reasons for holding it back. In that case, I'm interested in those reasons - it's always helpful to see another developer's train of thought.

Thanks in advance!

tlambert03:
I'm so sorry @ndxmrb you're not annoying at all.  You're incredibly patient and polite.  I'm just failing at being a maintainer right now (and still trying to settle my world and unbury myself after a cross atlantic move).  I will have an agent re review once more and just release it soon.  My apologies and thank you for the bump

PR #428, opened 2026-09-06
Preserve child event forwarding policy during replacement

### LINKED ISSUE #423: Signal group of dynamic child events bubble up even if connect_child_events is False
Events triggered on a child dataclass that is dynamically assigned to a parent dataclass are bubbled up to the parent even if `connect_child_events=False` is explicitly set on the parent's signal group.

This is true only if `parent.events.connect` is called _before_ the child is dynamically assigned.

## Details
If one evented dataclass contains another evented dataclass, then events bubble up by default. Setting `connect_child_events=False` on the parent's signal group disables event bubbling.

However, event bubbling is not correctly stopped if ALL of the following apply:
1. The parent and child _both_ have their own `SignalGroupDescriptor()`,
2. The child reference in the parent (`parent.child`) is mutated, AND
3. The listener (`parent.events.connect`) is registered _before_ the child reference is changed.

In other words, this bug does not occur if ANY of the following applies:
1. The child does not have its own `SignalGroupDescriptor()`,
2. The child reference in the parent (`parent.child`) is immutable, OR
3. The listener (`parent.events.connect`) is registered AFTER the child reference is changed.

## Reproducible code
```python
from dataclasses import dataclass

from psygnal import SignalGroupDescriptor


@dataclass
class Child:
    foo: int = 0

    events = SignalGroupDescriptor()


@dataclass
class Parent:
    child: Child | None = None

    events = SignalGroupDescriptor(connect_child_events=False)


if __name__ == "__main__":
    parent = Parent()
    parent.events.connect(lambda event: print(f"parent event: {event}"))

    child = Child()
    parent.child = child

    child.foo = 1
```

## Expected output
One line is printed to the console, namely when `parent.child` is assigned.

## Actual output
Two lines are printed: one when `parent.child` is assigned, and one when `child.foo` is assigned:

```
parent event: EmissionInfo(signal=<_DataclassFieldSignalInstance 'child' on <SignalGroup 'ParentSignalGroup' on <Parent instance at 0x23c858774d0> with 1 signals>>, args=(Child(foo=0), None), path=(.child,))
parent event: EmissionInfo(signal=<_DataclassFieldSignalInstance 'foo' on <SignalGroup 'ChildSignalGroup' on <Child instance at 0x23c85877620> with 1 signals>>, args=(1, 0), path=(.child, .foo))
```

## Details (cont.)
Returning to those conditions I mentioned earlier, you can verify them. If you do ANY of the following, you'll find that there is no `EmissionInfo(..., path=(.child, .foo)`, which is the expected behaviour:
1. Remove the child's `events = SignalGroupDescriptor()`,
2. Initialize the parent as `parent = Parent(Child())`, remove `child = Child()` and `parent.child = child`, and change `child.foo = 1` to `parent.child.foo = 1`, OR
3. Move `parent.events.connect(...)` to be right before `child.foo = 1`.

## Temporary workaround
In some cases, it may be sufficient to edit your listener(s) to explicitly check the `path` of the `event`.

Instead of
```python
parent.events.connect(lambda event: print(f"parent event: {event}"))
```

do
```python
def parent_listener(event):
    if len(event.path) != 1:
        return

    print(f"parent event: {event}")

parent.events.connect(parent_listener)
```

You'll have to do this for each listener separately, at all levels of your class hierarchy, which is of course undesirable.

## Environment details
**Python:** v3.14.6
**psygnal:** v0.15.1
**OS:** Windows 11 25H2

be-student:
Fixes #423.

Replacing an evented child currently connects it to the parent relay even when `connect_child_events=False`. Replacing a connected child with `None` also leaves its relay attached.

Track whether child forwarding has actually been enabled on each signal group, and maintain replacements only in that state. Keep explicit `connect_child_events()` opt-in working, including replacements before a listener is attached, and make repeated connection setup idempotent. Removing a child disconnects the old relay. Update the descriptor documentation and the signal-group type stub.

The regression matrix covers enabled/disabled forwarding, initial/absent children, early/late listeners, explicit opt-in, replacement, and removal. Existing direct-field child notifications and nested-container tests also pass.

Validation:
- Original regression: 6 failed, 2 passed before the fix.
- Full Python suite: 708 passed, 3 skipped.
- Built the mypyc package and verified compiled imports; full compiled suite: 707 passed, 4 skipped.
- Pinned Ruff checks and formatting pass.
- Mypy reports the same five errors on base and head (toolz typing, wrapt proxy signatures, and unavailable Qt stubs); logs compared identical. No Qt runtime was installed or exercised.


src/psygnal/_group.py
```diff
@@ -436,6 +436,7 @@ def __init__(self, instance: Any = None) -> None:
             for name, sig in cls._psygnal_signals.items()
         }
         self._psygnal_relay = SignalRelay(self._psygnal_instances, instance)
+        self._psygnal_child_events_connected = False
```
Czaki, line 439: Could you explain why you assign here the `False` value instead of depending on value passed in `SignalGroupDescriptor` like in issue?  `SignalGroupDescriptor(connect_child_events=False)`

src/psygnal/_group_descriptor.py
```diff
@@ -688,6 +696,7 @@ def __get__(
         if obj_id not in self._instance_map:
             # cache it
             self._instance_map[obj_id] = grp = signal_group(instance)
+            grp._psygnal_child_events_connected = self._connect_child_events
```
tlambert03: i believe this line is unnecessary for the fix? and might be the reason for the performance degradation test failure? (makes the "no listeners" fast path unreachable for every default `@evented` class)

```suggestion
```

src/psygnal/_group.py
tlambert03: Because `connect_child_events` and `_psygnal_child_events_connected` describe different states

  - `connect_child_events=True`: child forwarding is configured to be enabled when the first parent listener connects.

  - `_psygnal_child_events_connected=True`: child forwarding has already been established.
  
  i agree with this fix as is.

PR #433, opened 2026-09-11
fix: check if caller is alive then callback

### LINKED ISSUE #432: 0.16.0: emit crashes with ReferenceError when a weakly-connected object dies mid-emit
A signal calls its callbacks one by one. If one of those callbacks deletes the last reference to another connected object, psygnal 0.16.0 still calls that dead object's callback, and it raises `ReferenceError`. 0.15.1 skipped it.

```python
import gc
from psygnal import Signal

class Emitter:
    changed = Signal()

class Receiver:
    def __init__(self, drop=None):
        self.drop = drop
    def on_changed(self):
        if self.drop:
            self.drop()  # deletes the other receiver

emitter = Emitter()
holder = {}

first = Receiver(drop=lambda: holder.pop("second"))
second = Receiver()
holder["second"] = second

emitter.changed.connect(first.on_changed)
emitter.changed.connect(second.on_changed)
del second
gc.disable()  # refcounting only

emitter.changed.emit()
print("emit OK")
```

- psygnal 0.15.1: prints `emit OK`
- psygnal 0.16.0: `ReferenceError: weakly-referenced object no longer exists`, wrapped in an `EmitLoopError`

I think it comes from #413, which changed `for caller in self._slots` to `for caller in list(self._slots)`. The copy is what makes the self-disconnect case work, but it also keeps callbacks that were removed during the emit because their object was garbage collected. Those callbacks then get called anyway.

Where we hit it: napari's `TransformChain` weakly connects `self._clean_cache` to each of its child transforms. Setting a layer's `rotate` emits `changed`, one callback drops a chain, and the emit crashes on the next one. It broke our test suite through napari-orthogonal-views, so we pinned `psygnal<0.16` for now.

### Possible fix

`WeakCallback` already knows: `_alive` starts `True`, and `_kill_and_finalize` sets it to `False` the moment the weakref dies — the same place the slot is discarded. So the loop could check it:

```python
def _run_emit_loop_immediate(self) -> None:
    args = self._emit_queue.popleft()
    for caller in list(self._slots):
        if caller._alive:
            caller.cb(args)
```

I tried this against 0.16.0 (pure Python build): the snippet above prints `emit OK`, the self-disconnect behaviour from #413 still works, and your test suite passes (724 passed, 4 skipped).

`_run_emit_loop_latest_only` and `_run_emit_loop_queued` copy the list the same way and would need the same guard — I only patched the immediate one to test. You may also prefer the public `caller.dereference() is not None` over the private attribute.

Environment: Python 3.11–3.13, Linux/macOS/Windows.

TeunHuijben:
Fixed #432, see issue for details

### DIFF
--- a/src/psygnal/_queue.py
+++ b/src/psygnal/_queue.py
@@ -66,6 +66,9 @@ class QueuedCallback(WeakCallback):
     def dereference(self) -> Callable | None:
         return self._wrapped.dereference()
 
+    def _is_alive(self) -> bool:
+        return self._wrapped._is_alive()
+
     def __eq__(self, other: object) -> bool:
         """Compare QueuedCallback instances for equality based on wrapped callback.
 
--- a/src/psygnal/_signal.py
+++ b/src/psygnal/_signal.py
@@ -1336,8 +1336,14 @@ class SignalInstance:
 
     def _run_emit_loop_immediate(self) -> None:
         args = self._emit_queue.popleft()
+        # NOTE: iterating a copy of self._slots (so that a slot disconnecting
+        # itself doesn't skip the next one) means slots removed *during* this
+        # emit are still in hand. Skip the ones whose weakly-referenced object
+        # was garbage collected: they are already disconnected, and calling them
+        # raises ReferenceError.
         for caller in list(self._slots):
-            caller.cb(args)
+            if caller._alive:
+                caller.cb(args)
 
     def _run_emit_loop_latest_only(self) -> None:
         self._args = args = self._emit_queue.popleft()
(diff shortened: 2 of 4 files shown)


TeunHuijben:
Hi @tlambert03, I don't have the power to request reviews on thie repo, so hereby via message 😊

tlambert03:
thanks for the thorough issue and fix.  Sorry for the regression.  just sitting down to have a look now

tlambert03:
thanks very much @TeunHuijben.  I pushed a minor fix that ensures the liveness check also works for callbacks connected with `thread=...`, whose `QueuedCallback` wrapper otherwise retains a stale `_alive` value after the wrapped receiver is collected.  (along with a test for that case)

Issue #421, opened 2026-06-10
`__file__` usage + mypyc leads to non-reproducible builds


mr-c:
Hello again,

The usage of `__file__` gets compiled in by mypyc, and in fact it is using the wrong path (the build path, not the post-installed path). Perhaps a more dynamic usage would work better?

Due to Debian's new reproducibility requirement, I will have to turn off mypyc compilation of psygnal until a fix can be found.

https://tests.reproducible-builds.org/debian/rb-pkg/unstable/amd64/psygnal.html
https://reproduce.debian.net/amd64/api/v1/builds/259993/artifacts/669198/diffoscope

Czaki:
Hm. I don't know of any other Pythonic way to determine file location other than `__file__` usage. Maybe you should fill out a bug report in mypyc as they should handle it properly during compilation?

tlambert03:
i believe #429 does indeed fix this (debug_level=0 produces byte-identical wheels, from the `hatch-mypyc` and `mypyc` source: it only drops the C compiler's debug symbols (-g1 → -g0), with no effect at runtime)

mr-c:
Thank you @tlambert03 for your attention; unfortunately for Debian we do need the debug information, so that workaround will have to be reversed on our side.

mr-c:
> Hm. I don't know of any other Pythonic way to determine file location other than `__file__` usage.

https://docs.python.org/3/library/importlib.resources.html is how I normally do it

tlambert03:
oh ... that's a shame.  I'm sorry I didn't wait for your response.  I will try to get back to this shortly and evaluate your suggestion as an alternate fix

tlambert03:
*edit: this is claude, not me*

Confirmed and dug into this further — writing it up here so it's captured for whoever hits it next.

## Root cause

It's not `__file__` at runtime — it's mypyc's build step embedding the **compile-time build path** into the compiled `.so`'s debug info. hatch-mypyc builds inside `tempfile.TemporaryDirectory()` by default, which gets a fresh random name every build, and that path ends up baked into `.debug_line`/STABS/`gnu_debuglink` data (confirmed via mr-c's diffoscope: `/tmp/tmpu7v23wzw/build/__native_....c` vs `/tmp/tmp7zobu0o5/build/__native_....c`).

## Why #429's `debug_level=0` wasn't the right fix

[...]

PR #434, opened 2026-09-11
build: revert debug_level=0, document the real reproducible-builds fix

tlambert03:
Reopens #421 — mr-c [confirmed](https://github.com/pyapp-kit/psygnal/issues/421#issuecomment-5634910614) that #429's fix (`debug_level = "0"`) strips C debug symbols entirely, which Debian's `-dbgsym` packaging needs and will have to explicitly revert on their side.

## What this does

Reverts `options = { debug_level = "0" }`, restoring mypyc's default (debug info present), and replaces it with a comment explaining the actual fix and why `debug_level=0` shouldn't be reintroduced.

## Why this is safe / what the real problem was

The non-reproducibility isn't from disabling `-g0` — it's from hatch-mypyc building inside a fresh `tempfile.TemporaryDirectory()` every time, so the random path gets embedded in the compiled `.so`'s debug info. `debug_level=0` "fixed" it only by removing debug info altogether.

Debian's own packaging (`debian/rules`, added in 0.15.1-3 by Colin Watson, independent of anything in this repo) fixes the actual cause instead: it points `HATCH_MYPYC_BUILD_DIR` at a fixed, pre-created directory, combined with dpkg's default `-ffile-prefix-map`. [tests.reproducible-builds.org currently shows psygnal as reproducible](https://tests.reproducible-builds.org/debian/rb-pkg/unstable/amd64/psygnal.html) as a result — with debug info intact.

I verified this locally (macOS, default `debug_level`): two separate builds from the same project root, with `HATCH_MYPYC_BUILD_DIR` fixed and pre-created, embed byte-identical paths in the compiled `.so`. Full writeup with the details in [the issue comment](https://github.com/pyapp-kit/psygnal/issues/421).

## Verification
- Compiled build (mypyc, default debug_level) on py3.14: 735 passed, 8 skipped
- `prek run --files pyproject.toml`: passed
- Wheel size goes back up by the ~11% #429 saved (debug info restored) — no other size/behavior impact

## Not included here
Wiring `HATCH_MYPYC_BUILD_DIR` into our own release CI so our published wheels are byte-reproducible too. cibuildwheel builds each platform in its own container/venv and I couldn't verify the container path is stable across runs from here, so I left it as a documented follow-up in the issue rather than shipping something untested.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_011J3gAnJNwqyWvosDvLwNPE

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -100,7 +100,18 @@ sources = ["src"]
 
 [tool.hatch.build.targets.wheel.hooks.mypyc]
 mypy-args = ["--ignore-missing-imports"]
-options = { debug_level = "0" }
+# NOTE: do not set `options.debug_level = "0"` here to "fix" reproducible builds
+# (see #421/#429). It works, but it strips C debug symbols from every wheel we
+# ship, which downstream packagers who build -dbgsym packages (e.g. Debian) need
+# and have to explicitly restore. The actual source of non-determinism is
+# hatch-mypyc's default `tempfile.TemporaryDirectory()` build root, which embeds
+# a random path into the compiled .so's debug info on every build. Debian's own
+# packaging already fixes this correctly, without touching debug_level, by
+# pointing `HATCH_MYPYC_BUILD_DIR` at a fixed, pre-created directory instead of
+# the random default (see debian/rules), combined with dpkg's default
+# `-ffile-prefix-map` (the `fixfilepath` build flag) to normalize the remaining
+# absolute-build-root prefix. Confirmed locally: with a fixed `build-dir` alone
+# (no debug_level change), two separate builds embed byte-identical paths.
 enable-by-default = false
 require-runtime-dependencies = true
 dependencies = [


tlambert03:
@mr-c does is there any way for you to confirm if this would work better for you/debian?

I.e: you can fix it yourself using `HATCH_MYPYC_BUILD_DIR` (provided we revert the PR that stripped debug that is)

Or: would you prefer that we really do look into never using things like `__file__` instead

mr-c:
@tlambert03 Turns out that Colin Watson already fixed the problem, sorry for the noise. Yes, please don't strip out the debug data!

I find that `__file__` tends to cause other problems; but I can't give you a specific example for psygnal off-hand

tlambert03:
> I find that __file__ tends to cause other problems

ok... well, if we're "good enough" for now and nothing is actively burning, let's try to go with it for now.  Please don't hesitate to open an issue if you do find that `__file__` is definitely causing issues and I'll fix it then

tlambert03:
merged.   i can't quite remember if it's enough to just merge this into main for your needs... but please just let me know if you need a new tag or release.

PR #435, opened 2026-09-12
docs: Fix and complete numpydoc Parameters sections

xuanku-001:
## Summary

Fixes and completes numpydoc `Parameters` sections in the container and event APIs:

- `connect_child_events`: the section header was written as `Parameters.` (with a
  trailing period), which numpydoc does not recognize, so the parameter list below it
  was rendered as plain text.
- `qt.stop_emitting_from_queue`: the parameter list (`thread`) had no `Parameters`
  header at all.
- `SelectableEventedList.select_previous`: `select_next` right above it documents
  `expand_selection` and `wraparound`; `select_previous` takes the same arguments but
  had no section. The new section mirrors the existing one.
- `SelectableEventedList.insert`, `EventedList.reverse`, `SignalRelay.block`, and
  `Selectable.selection`: their parameters were not described.

All existing sentences are preserved verbatim; only sections are added, so no
behaviour or signatures change. These docstrings are rendered on the docs site with
`docstring_style: numpy`, so the two header fixes directly affect the rendered API
reference (the previously broken lists are currently shown as plain text).

### Testing

- `python -m py_compile` passes on all touched files
- AST statement counts are unchanged (documentation only, no executable lines added)
- All parameter names cross-checked against signatures (ruff `D417`-consistent), and
  the section headers follow the existing `Parameters` / `----------` pattern used
  elsewhere in the codebase

---
*This contribution was prepared with AI assistance (a local LLM-based agent harness with deterministic checks and human review).*


src/psygnal/containers/_selection.py
```diff
@@ -191,6 +191,12 @@ def selection(self) -> Selection[_S]:  # pragma: no cover
 
     @selection.setter
     def selection(self, new_selection: Iterable[_S]) -> None:  # pragma: no cover
-        """Set selection, without deleting selection model object."""
+        """Set selection, without deleting selection model object.
+
+        Parameters
+        ----------
+        new_selection : Iterable[_S]
+        The items to select.
```
tlambert03:
```suggestion
        new_selection : Iterable[_S]
            The items to select.
```

